PR #8, opened 2019-07-23
add test coverage

Thomaash:
At the moment:
- Updated dependencies.
- Coverage works (like 40 % is touched by test).
- Tests pass.
- Project builds.
- Examples work.
- There is one new test (for DOT parser).
- TypeScript is ready to be used (the new test is in TS).

More changes will follow.


Thomaash:
It introduces TypeScript support and tests written in TypeScript. But okay, there is no TS code in the code base yet.

Issue #13, opened 2019-07-25
Migrate DefinitelyTyped typings for TypeScript support


jamesorlakin:
There was a type declaration file on DefinitelyTyped for vis to allow TypeScript developers to enable static analysis and autocomplete in VS Code. Would this project be interested in migrating it over for vis-network?

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/vis/index.d.ts

I would happily give it a go but the file was never that accurate and I'm not that concrete on vis' inner workings.

gflarity:
@jamesorlakin I'll definitely use it and I'm more than willing to fix and issues I run into it. Might as well go for it?

gflarity:
There's also a tsconfig.json in this project, but no .ts files... so maybe there's something in works?

mojoaxel:
> There's also a tsconfig.json in this project, but no .ts files... so maybe there's something in works?

Thanks to @Thomaash the whole visjs family is moving into the typescript direction :tada: 
The vis-network repository already has one test file written in TS and I'm sure @Thomaash is planning to move a lot in this direction in the future.

Before we change the whole codebase to typescript We really should try to adopt as many existing pull-requests from [almende/vis](https://github.com/almende/vis/pulls) and [visjs-community/visjs-network](https://github.com/visjs-community/visjs-network/pulls?q=is%3Apr+is%3Aclosed) as possible!

Thomaash:
There are actually 2 tests in TypeScript (that's why the tsconfig.json exists). I wan't to rewrite it into TypeScript (DataSet and DataView already are) but I also want to be sure to not break anything (or break as few things as possible). However there are very few tests to help me with this, so my plan right now is to increase the test coverage first.

I originally started with visjs-network (it seemed more active to me) but that didn't go that well. It just lies on my HDD and it's probably not worth publishing. I just jumped into it, made some big breaking changes and then realized they're not only incompatible but also nonsensical (quite a lot of wasted effort). I don't want to do this mistake again.

mojoaxel:
>  I don't want to do this mistake again.

:+1: We all lern all the time! I would recommending starting small and than moving more and more to TS. As I understand it JS is compatible with TS so the first step could be to rename the first file to `.ts` and start adding type information!?

At some point https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/vis could become obsolete.

glennverschooren:
Hi, 

[...]

PR #20, opened 2019-07-26
Use TypeScript index file

Thomaash:
This reexports types from other packages (vis-util and vis-data).
**We have no types here so local files point nowhere.**

Partially addresses #13.


rollup.config.js
```diff
@@ -1,32 +1,69 @@
-import commonjs from 'rollup-plugin-commonjs';
-import nodeResolve from 'rollup-plugin-node-resolve';
-import nodeBuiltins from 'rollup-plugin-node-builtins';
-import babel from 'rollup-plugin-babel';
-import sourcemaps from 'rollup-plugin-sourcemaps';
-import { uglify } from "rollup-plugin-uglify";
-import banner from 'rollup-plugin-banner';
-import genHeader from './lib/header';
+import commonjs from 'rollup-plugin-commonjs'
```
mojoaxel: please don't change existing code if you don't have to.
Cody style changes ALWAYS should be done in a seperate pull-request.

rollup.config.js
```diff
@@ -1,32 +1,69 @@
...
+  ...plugins.slice(0, 4),
+  minify({ comments: false }),
+  ...plugins.slice(4)
+]
+
+export default [
+  {
+    input: 'lib/index.ts',
+    output: {
+      file: 'dist/esm.js',
```
mojoaxel: see above..

rollup.config.js
```diff
@@ -1,32 +1,69 @@
...
+      format: 'umd',
+      exports: 'named',
+      name: 'vis',
+    },
+    plugins,
+  },
+  {
+    input: 'lib/index.ts',
+    output: {
+      file: 'dist/esm.min.js',
```
mojoaxel: see above..

mojoaxel:
@Thomaash Thanks! I'm really looking forward to this!

rollup.config.js
Thomaash: The files are too different, Git just doesn't see it.

rollup.config.js
Thomaash: The files are too different, Git just doesn't see it.

package.json
```diff
@@ -23,15 +23,18 @@
     "network",
     "browser"
   ],
-  "main": "./dist/vis-network.min.js",
```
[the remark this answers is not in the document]
Thomaash: I don't know. It seems pretty straight forward to me that `node_modules/vis-network/dist/umd.min.js` is the UMD build of vis-network package.

And why would it be a major change? You import the package, the package.json has main that points to a file, the path of the file doesn't matter.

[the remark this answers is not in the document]
Thomaash: Also we have it this way in vis-util, vis-uuid and vis-data. If it's too cryptic and a major change we should discuss it there too.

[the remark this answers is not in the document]
Thomaash: Oh, it breaks everything **inside** of this package :smile:.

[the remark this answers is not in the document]
mojoaxel: The problem is, that not everybody used a package manager. In a lot if use-cases people still just integrate a script tag with the library they need. So if the file `dist/vis-network.min.js` suddenly does't exist anymore than this is breaking change.

I have no problem with this technical name in the `vis-uuid`, `vis-util` and even `vis-data` package because thay are basically just helper packages that are propably never integrated in a website.

I still would ask you to keep the name (at least for now).

types/network/Network.d.ts
```diff
@@ -0,0 +1,1049 @@
+// Type definitions for vis.js 4.21
```
mojoaxel: Here the version is no longer correct and could be removed.

Also i'm not sure if we should just copy this file into this repository!? But what are the alternatives?
- We could cherry-pick the network commits into this repository. This ways we would keep the git history of the file, but this would be really messy...( Maybe I'll try it! It's a challange! )
- We could just fork the `DefinitelyTyped` repository and maintain the types there. But this would be hard to keep up to date. I like the idea to have the types inside the repository...

mojoaxel:
@Thomaash I would prefer it to start really only with the index.ts und than move more and more components over to TS in separated (reviewable) pull-requests. This thing here is becoming a reals monster :-)

Thomaash:
@mojoaxel Okay, so if I understand correctly:
1. PR with index.ts (disfunctional types, but we don't care since we don't have functional anyway).
1. PR with DefinitelyTyped stuff (making the types at least somewhat usable).
1. PR with TypeScript Gephi parser.

Seems reasonable to me.

mojoaxel:
> PR with index.ts (disfunctional types, but we don't care since we don't have functional anyway).

Yes, I would say just the TS dependencies, builds and maybe one proove-of-concept test file (Oh wait we already have this :smile: )

> PR with DefinitelyTyped stuff (making the types at least somewhat usable).

I'm also looking into this. I'm trying to move the existing files from `DefinitelyTyped` over without loosing the git history.

> PR with TypeScript Gephi parser.

That could be a good start! It's important, that the pull-request stay small enough to be able to review them. I also want to attract other maintainer and small and well structured PRs are a good way to do that :+1: 

Sounds good!

Thomaash:
@mojoaxel This is now only about index.js to index.ts transition and adjusting the build process to this change.

lib/header.js
```diff
@@ -32,11 +32,13 @@ vis.js may be distributed under either license.`;
...
  */
 function genHeader(component) {
-	return [
+   return '/*\n' + [
       pkg.name,
       component ? ' - ' + component : '',
       header
-   ].join('');
+   ].join('')
+   .replace(/^(?!\n)/gm, ' * ')
```
mojoaxel: I would prefer it if you would leave the `header.js` untouched for now. I'm not sure if we want the same copyright-header an every visjs module in the future.

Also one could argue if this empty spaces should be there or not...Let's just not discuss this, yet. I think we have more important things to do :rofl:

Thomaash: Okay, I'll move it to rollup.config.js.

mojoaxel: > Okay, I'll move it to rollup.config.js.

Its not about the `header.js` file, but the header it produces. For now it is the same header overall projects. Maybe this is a good thing!?
Why do you want to change it? Ist eslint unhappy?

Thomaash: The problem here is that rollup-plugin-banner messes up source maps. Rollup has native support for banners (headers) but they have to be valid JavaScript comments. This code just turns it into a comment.

Thomaash: BTW: The final header looks the same.

Issue #7, opened 2019-07-23
adopt pull-requests fom vis and visjs-timeline


mojoaxel:
We should think about how to adopt (old) pull-requests from [vis](https://github.com/almende/vis/pulls?q=is%3Aopen+is%3Apr) and [visjs-network](https://github.com/visjs-community/visjs-network/pulls?q=is%3Apr+is%3Aclosed).

- [x] [almende/vis#4039](https://github.com/almende/vis/pull/4039), [visjs-network/pull/6](https://github.com/visjs-community/visjs-network/pull/6) and [visjs-network/pull/9](https://github.com/visjs-community/visjs-network/pull/9)
- [x] [almende/vis/pull/3963](https://github.com/almende/vis/pull/3963), [visjs-network/pull/20](https://github.com/visjs-community/visjs-network/pull/20)
- [x] [almende/vis#3888](https://github.com/almende/vis/pull/3888), [visjs-network/pull/23](https://github.com/visjs-community/visjs-network/pull/23)
- [x] [almende/vis#4026](https://github.com/almende/vis/pull/4026), [visjs-network/pull/26](https://github.com/visjs-community/visjs-network/pull/26)
- [x] [almende/vis#4027](https://github.com/almende/vis/pull/4027), [visjs-network/pull/29](https://github.com/visjs-community/visjs-network/pull/29)
- [x] [almende/vis#3771](https://github.com/almende/vis/pull/3771),  [visjs-network/pull/34](https://github.com/visjs-community/visjs-network/pull/34)
- [x] [visjs-network/pull/39](https://github.com/visjs-community/visjs-network/pull/39)
- [x] [almende/vis#3657](https://github.com/almende/vis/pull/3657) and [visjs-network/pull/40](https://github.com/visjs-community/visjs-network/pull/40)
- [x] [visjs-network/pull/43](https://github.com/visjs-community/visjs-network/pull/43)
- [x] [visjs-network/pull/45](https://github.com/visjs-community/visjs-network/pull/45)
- [x] [visjs-network/pull/47](https://github.com/visjs-community/visjs-network/pull/47)
- [x] [visjs-network/pull/66](https://github.com/visjs-community/visjs-network/pull/66)
- [x] [visjs-network/pull/68](https://github.com/visjs-community/visjs-network/pull/68)
- [x] [visjs-network/pull/71](https://github.com/visjs-community/visjs-network/pull/71)

[...]

PR #25, opened 2019-07-26
visjs-network#40: fix type of zoomSpeed

mojoaxel:
This is an original contribution from @geminoa: https://github.com/visjs-community/visjs-network/pull/40

this is an addition to https://github.com/almende/vis/pull/3657 that is already merge here.

part of #7


mojoaxel:
This solved the following error:
image

PR #35, opened 2019-07-28
chore(test): added visual regression tests

### LINKED ISSUE #33: create automated regression tests
We should introduce screenshot/regression tests!
I would recommend https://www.cypress.io

mojoaxel:
fixes #33


Thomaash:
Oh, I have to start the server manually.

Thomaash:
Are you sure the nodes are always placed exactly the same? All the tests fail for me with what seems as slight misalignment of nodes.

Thomaash:
nodeStyles - HTMLInNodes -- match screenshot diff

mojoaxel:
> Are you sure the nodes are always placed exactly the same? All the tests fail for me with what seems as slight misalignment of nodes.

Oh no!
@Thomaash  can you please try `npm run test:cypress`

mojoaxel:
> Oh, I have to start the server manually

Yeah :disappointed: We need a solution for this...

Thomaash:
Try this: examples/network/layout/randomSeed.html .

Thomaash:
If it works we could simple set all the examples the same way. Or maybe all but one to show the default behavior.

Thomaash:
Pulled the recent changes: 12/12 fail.

mojoaxel:
> All the tests fail for me with what seems as slight misalignment of nodes.

Yes the rendering depends on the browser and the OS, installed fonts... :disappointed: 
Even if we are not able to fix this we maybe still could use it from inside Travis because it has a reliable environment!?

Thomaash:
We can fix the environment to specific version so it should work.

Thomaash:
We can ship some default fonts, use hardcoded seed in the examples. AFAIK Cypress packs it's own electron so even this should be consistent (hopefully) if we pin a specific version.

mojoaxel:
The tests are all broken ATM because of #60

mojoaxel:
closing in favor of #187 :tada:

Thomaash:
#187 doesn't replace this, it actually complements this. It would be best to have both.

PR #37, opened 2019-07-28
Add default export for backwards compatibility

Thomaash:
Fixes:
> There is no default export in ESM build (TBH this is my fault) therefore it cannot be imported as import vis from 'vis-network', it has to be imported as import * as vis from 'vis-network'. Although UMD build works (hence working examples). This is definitely breaking change because tools like Webpack or Rollup automatically pick ESM over UMD.


Thomaash:
You're welcome.

PR #39, opened 2019-07-28
Rewrite Gephi parser into TypeScript

Thomaash:
Apart from obvious this also includes a bugfix: `options.edges.inheritColor` was originally read as `options.inheritColor` and therefore ignored.


Thomaash:
I tried it. `git mv` works from one commit to another if there are not too many changes. Across multiple commits or with many changes (like types inserted almost everywhere) it just fails.

mojoaxel:
> I tried it. git mv works from one commit to another if there are not too many changes. 

But could you not just move/rename the file than commit it and than start changing using new commits? (I'f never done that. I'm just curious!)

Thomaash:
I did it. It works if you go commit by commit. If you want to see the entire thing it breaks anyway.

mojoaxel:
> If you want to see the entire thing it breaks anyway.

But without the `git mv` the history of this file gets lost :-(

Thomaash:
But with it too. I tried to retain the history, Git just loses track once the changes are too big. And they will be too big if you squash this PR.

mojoaxel:
Thanks for clarifying!
Yes this squashing is a real destroyer of the git history...well it is how it is!
Thanks for trying!

Thomaash:
No problem.

PR #44, opened 2019-07-29
fix: ensure that invalid scale is not passed to moveTo

### LINKED ISSUE #12: bad rendering and network.body.view.translation is {x:NaN, y:NaN}
Hi

There is a hard-to-reproduce bug with vis network 4.21.0 : sometimes, the networks is not rendered properly (only edges), and it is related to network.body.view.translation being {x:NaN, y:NaN}.

I'm getting the issue on all kind of browser and on different computers, especially the slower one. Maybe the browser is trying to render before the network has been initialized ?

I've found I can reset the view translation to x:0,y:0, then redraw().

Is this bug solved in the 5.0 ?

Regards

Thomaash:
Coerce to number and throw if it isn't grater than zero.

Fixes #12.

All MWE from the issue either work correctly or throw right away.

### DIFF
--- a/lib/network/modules/View.js
+++ b/lib/network/modules/View.js
@@ -142,7 +142,7 @@ class View {
     // Coerce and verify that the scale is valid.
     options.scale = +options.scale
     if (!(options.scale > 0)) {
-      throw new TypeError('Scale has to be a number bigger than zero.')
+      throw new TypeError('The option "scale" has to be a number greater than zero.')
     }
 
     if (options.offset    === undefined)           {options.offset    = {x: 0, y: 0};    }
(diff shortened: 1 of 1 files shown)


lib/network/modules/View.js
```diff
@@ -138,6 +138,13 @@ class View {
       options = {};
       return;
     }
+
+    // Coerce and verify that the scale is valid.
+    options.scale = +options.scale
+    if (!(options.scale > 0)) {
+      throw new TypeError('Scale has to be a number bigger than zero.')
```
mojoaxel: How about? `'scale' has to be a number bigger than zero.`

Thomaash: Probably a bit more explicit. Will do.

Issue #1, opened 2019-07-17
introduce semantic-release


mojoaxel:
https://www.npmjs.com/package/semantic-release

mojoaxel:
I'm not sure if `semantic-release` can do this but I really would love to auto-release after each successful pull-request.

Thomaash:
Resolved by #48.

PR #48, opened 2019-07-31
chore(CI): add CircleCI config

Thomaash:
Installs, tests, builds and releases (semantic-release) if necessary.
Linting is commented out at the moment (too many failures). It should be uncommented later when the code base is compliant.

Addresses #1.

PS @mojoaxel: I can't set this (or any other repo) up on CircleCI. It says: “You must be an admin to add a project on CircleCI”.

### DIFF
--- a/.releaserc.yml
+++ b/.releaserc.yml
@@ -1,0 +1,6 @@
+branch: master
+plugins:
+  - '@semantic-release/commit-analyzer'
+  - '@semantic-release/release-notes-generator'
+  - '@semantic-release/npm'
+  - '@semantic-release/github'
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "vis-network",
-  "version": "5.1.1",
+  "version": "0.0.0-no-version",
   "description": "A dynamic, browser-based visualization library.",
   "homepage": "http://visjs.org/",
   "license": "(Apache-2.0 OR MIT)",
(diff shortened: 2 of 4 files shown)


mojoaxel:
> I can't set this (or any other repo) up on CircleCI. It says: “You must be an admin to add a project on CircleCI”.

@Thomaash You are now also a admin of this repository. Please be very responsible! (One time a contributor deleted the whole almende/vis repository from github, by mistake.)

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "vis-network",
-  "version": "5.1.1",
+  "version": "0.0.0-no-version",
```
mojoaxel: why this?

.releaserc.yml
```diff
@@ -0,0 +1,6 @@
+branch: master
+plugins:
+  - '@semantic-release/commit-analyzer'
```
mojoaxel: do we need this packages as dependency in package.json?

Thomaash:
Thanks, I'll be very careful.

.releaserc.yml
Thomaash: Nope, semantic-release depends on these.

package.json
Thomaash: We will release through semantic-release which doesn't update this. In other words we would have 5.1.1 forever.

Issue #65, opened 2019-08-11
bundle css with rollup


mojoaxel:
CSS bundling should be done with rollup (like e.g. in the timeline).
At the moment `vis-charts` completely ignores vis-timeline styles and don't bundles them.

Thomaash:
Done.

mojoaxel:
fixed via #69

PR #69, opened 2019-08-13
Use Rollup to bundle CSS

Thomaash:
The CSS files are now imported from within JS files.

Note: CSS Porter doesn't seem to be able to generate source maps, therefore there are no source maps for CSS files.

Addresses #65.


package.json
```diff
@@ -51,7 +49,7 @@
     "moment": "^2.24.0",
     "timsort": "^0.3.0",
     "vis-data": "^6.1.1",
-    "vis-util": "^1.1.2"
+    "vis-util": "^1.1.4"
```
mojoaxel: dependency update should better be done it a separat PR if it is not related.

Thomaash: Done.

PR #77, opened 2019-08-14
fix(layout-engine): invalid hierarchical layout

### LINKED ISSUE #4: hierarchical directed layout: wrongly directed arrows
_The fiddle below is using vis.js as vis-network seems to be not yet available on cdnjs, however this is not working on the current vis-network version either._

As you can see on [this jsfiddle](http://jsfiddle.net/Ln4t9mjr/) not strictly hierarchical networks do not very well with vis atm. It seems that the levels given to the diferent nodes by vis are not quite.. right.
The first network shows nodes without the level argument given, the second one below with level given. Without explicit level the result is just wrong: crossing edges and even wrong levels. (In up direction a child node should never be above its parent node ^^) 

The edge crossing seems to be [known](https://github.com/almende/vis/issues/3357) but it still would be nice if it gets fixed or at least if one could toggle if he want "optimazation but crossing" or "no crossing but no optimization" mentioned [here](https://github.com/almende/vis/issues/3357#issuecomment-322148978).
Concerning the wrong direction for some nodes I did not find any related issue.

Im not sure if it can help here, but it seems like for larger graphs without level given the order of edges does alter the resulting graph significantly. (I will update the issue with an example as soon as I got one or maybe better open another one, as it may be a totally different thing.) 

Without having looked into the code, I assume the graph engine isnt sorting the edges it is given properly in order to assign them their final level in the graph, thus nodes are placed on wrong levels resulting in wrongly directed arrows/edges.

Thomaash:
Creates valid hierarchy without edges sometimes leading to the upper levels and other weirdness like nondeterministic behavior (cyclic hierarchies still use the flawed logic but it's impossible to organize a cycle into a hierarchy so there's no better solution anyway).

Performance (`L` is the amount of leaf nodes, `N` the amount of nodes and `E` the amount of edges):
- On a balanced binary tree with 10 000 nodes it visited only 10 040 nodes (basically `O(N)`), but that's the best case scenario.
- The worst case will be around `O(L*N*E)`.

MWEs:
- Before: http://jsfiddle.net/h3n91rme/
- After: http://jsfiddle.net/oae35sk7/

Resolves #4.


Thomaash:
MWE: http://jsfiddle.net/sat61b72/ (on top how it works, bottom how it should work).

lib/network/modules/components/DirectionStrategy.js
```diff
@@ -3,7 +3,8 @@
  *
  * Strategy pattern for usage of direction methods for hierarchical layouts.
  */
-import TimSort from 'timsort';
+
+import { sort as timsort } from 'timsort';
```
Thomaash: This isn't really obvious but Timsort has no default export. It works in Rollup since it's UMD. It doesn't however work in Mocha. This way it works in both build and test.

rollup.config.js
```diff
@@ -16,7 +16,11 @@ const plugins = {
 	typescript: typescript({
 		tsconfig: 'tsconfig.code.json'
 	}),
-	commonjs: commonjs(),
+	commonjs: commonjs({
+		namedExports: {
+			'timsort': ['sort']
+		}
```
Thomaash: Timsort exports sort but Rollup fails to detect this so I had to set it here manually.

Thomaash:
Linting fails because @typescript-eslint/member-ordering rule doesn't work well, disabled through #93.

Thomaash:
I don't think this will break anything. The hierarchical layout is broken and this fixes it (at least it improves it). Who would rely on the layout being messed up? I'd go for bug fix release.

PR #88, opened 2019-08-23
style(prettier): enforce default style in new TS files

Thomaash:


.eslintrc.js
```diff
@@ -102,5 +102,88 @@ module.exports = {
...
+        "test/edges/helpers.ts",
+        "test/edges/straight-edge.test.ts",
+        "test/gephi-parser.test.ts",
+        "test/helpers/index.ts"
+      ],
+      rules: {
+        "prettier/prettier": [
+          "error",
+          {
+            endOfLine: "lf",
```
mojoaxel: @Thomaash I thought we decided to use the default prettier config?

.eslintrc.js
```diff
@@ -102,5 +102,88 @@ module.exports = {
         'valid-typeof': 'off',
       },
     },
-  ],
-}
+    {
+      files: [
+        "lib/index-bundle.ts",
```
mojoaxel: could we not use a glob like `lib/**/*.ts`?

.eslintrc.js
```diff
@@ -102,5 +102,88 @@ module.exports = {
...
+            requireParamDescription: true,
+            requireParamType: false,
+            requireReturn: false, // Requires return for void functions.
+            requireReturnDescription: true,
+            requireReturnType: false
+          }
+        ],
+
+        // Class related.
+        "@typescript-eslint/member-naming": [
```
mojoaxel: Once more: Why do we need eslint if we already have prettier?

Thomaash: Prettier is only about formatting. ESLint does a lot more than that (it reports unused variables, missing JSDoc etc.)

.eslintrc.js
Thomaash: This is just for existing files. The default config will be enforced in new files.

.eslintrc.js
Thomaash: If we reformat the old files, then yes.

mojoaxel: Ok, this should be cleaned up later, than.

.eslintrc.js
mojoaxel: I think we should leave th old *.js files alone and just format the new *.ts files.

Thomaash: Eh, this is exactly what this config does.

PR #95, opened 2019-08-28
ci: update GitHub Pages from CI

Thomaash:
Automatically updates gh-pages branch on every master commit. Solves issues mainly with dist creating unnecessary merge conflicts or getting outdated because someone forgot to rebuild it.

Closes #152.


.gitignore
```diff
@@ -18,3 +18,4 @@ gen/
 coverage/
 .rpt2_cache
 vis-network-*.tgz
+/dist/
```
mojoaxel: I think we talked about this earlier. I'm not sure if we simply can remove the `dist`folder because apps may depend on it.

.circleci/config.yml
```diff
@@ -82,6 +82,25 @@ jobs:
...
+          name: Set Git credentials
+          command: |
+            echo 'machine github.com' >> ~/.netrc
+            echo "        login $GIT_AUTHOR_NAME" >> ~/.netrc
+            echo "        password $GH_TOKEN" >> ~/.netrc
+
+      - run:
+          name: Publish GitHub Pages
+          command: |
+            npx gh-pages --dist . --message "chore: update to $(git rev-parse HEAD) [ci skip]" --repo "$(node -e 'process.stdout.write(require("./package.json").repository.url)')" --src '{{coverage,dist,docs,examples}/**/*,LICENSE*,*.{html,md}}'
```
mojoaxel: We are currently not using the gh-pages branch but use master for the github-pages feature. This ways ew do not have wo sync two branches and can always be sure the examples are using the latest master version (e.g. useful for bug-reports).

I really would like to keep it that way!

.gitignore
Thomaash: We are not removing it, we are moving it to gh-pages branch and publishing that instead. Nothing changes from the outside we'll just switch GitHub Pages from master to a branch.

.circleci/config.yml
Thomaash: As long as everybody always updates everything manually, sure. But what if someone doesn't rebuild dist in PR? That can happen and already happened. This way everything would happen automatically in CI.

mojoaxel: Sounds reasonable!
I'll try to have a look this week and try to figure out if this breaks something...

mojoaxel:
## :+1: Pros
- The dist files would no longer show up in the PR and would not need to be ignored anymore in reviews.
- The docs would always fit the released sources (I think this is still the case here due to the semantic-releases).
- Merges would no longer include the dist files (easier merges)

## :-1: Cons
- The dist files would no longer be includes in the [github released archives (zip, tar.gz)](https://github.com/visjs/vis-network/releases).
- Contributors may think the need to change the `gh-pages` branch if they want to fix e.g. a typo. (This actually was the case with `alemende/vis).
- Running code that links to the existing `dist` files (e.g. from jsbin using rawgit) would no longer work.
- More complexity

:thinking: I'm still not sure if it's worth it. I made the conscious decision to use only the master branch for source and documentation. I did this to reduce complexity. Of course this has it's downsides but I see that we increase the complexity more and more - this makes it harder and harder for new maintainers to join.

@Thomaash One again thanks again for your work but I'm not willing to Approve this alone. 
**We should discuss this a little bit in the community!**

Thomaash:
Existing links through GitHack etc. will indeed stop working. However instead of linking the dist files through GitHack you can link the GitHub Pages (for example https://visjs.github.io/vis-network/dist/vis-network.js). This method already works and will work even if this is merged.

Thomaash:
Side note: RawGit will shut down soon anyway :smile:.

Thomaash:
I bit messier than this but solving your cons would be to commit and push back into the repo from CI (https://support.circleci.com/hc/en-us/articles/360018860473-How-to-push-a-commit-back-to-the-same-repository-as-part-of-the-CircleCI-job). This way we can tell people to not commit dist and other generated files (but if they do it won't be a problem) and be sure that everything is up to date.

Thomaash:
Hi @mojoaxel,

could you have a look at this, please? It's now in line with https://github.com/visjs/vis-timeline/pull/139 and https://github.com/visjs/vis-timeline/pull/146 which together work without issues.

PR #112, opened 2019-09-03
fix(node-color): prevent color from being overwritten when value is also set

### LINKED ISSUE #111: overriding defaultGroups color
Overriding defaultGroups background color does not work when at least one node.value is set. 

[example](https://jsbin.com/sihuwag/edit?html,output)

Thomaash:
Closes #111.


Thomaash:
@mojoaxel I just had to rebuild dist to resolve merge conflicts.

PR #115, opened 2019-09-04
fix(layout): update layout when an edge is reconnected

### LINKED ISSUE #114: Updating edges ends in hierarchical layout doesn't reset the layout
If one decides to update the `from` or `to` property of an edge, the hierarchical layout might have to be recomputed. This behavior is currently not supported.
Here is an example:
https://jsfiddle.net/57j0pugd/
Just click on the update button and you'll see the behavior. Calling `setupHierarchicalLayout` fixes it after a redraw.

My understanding of the problem is that `Edge.setOptions()` doesn't return `true` when `from` or `to` is changed. So this call in `EdgesHandler.update()` at (1) returns `false`:
```javascript
  update(ids) {
    var edges = this.body.edges;
    var edgesData = this.body.data.edges;
    var dataChanged = false;
    for (var i = 0; i < ids.length; i++) {
      var id = ids[i];
      var data = edgesData.get(id);
      var edge = edges[id];
      if (edge !== undefined) {
        // update edge
        edge.disconnect();
// (1)
        dataChanged = edge.setOptions(data) || dataChanged; // if a support node is added, data can be changed.
        edge.connect();
      }
      else {
        // create edge
        this.body.edges[id] = this.create(data);
        dataChanged = true;
      }
    }

    if (dataChanged === true) {
      this.body.emitter.emit('_adjustEdgesForHierarchicalLayout');
      this.body.emitter.emit("_dataChanged");
    }
    else {
      this.body.emitter.emit("_dataUpdated");
    }
  }
```
Then it only emits '_dataUpdated' which doesn't trigger a call to `LayoutEngine.setupHierarchicalLayout()`, which I assume is the function that would reset the layout.

So we could add a listener on '_dataUpdated' in `LayoutEngine`, but I think the problem could also come from `Edge.setOptions()`. Shouldn't a change to the `from` or `to` property of an edge be considered a change to the data, and `Edge.setOptions()` return `true` in this case ?

Thomaash:
Before: https://jsfiddle.net/57j0pugd/.
After: https://jsfiddle.net/thomaash/1rn6eudb/.

Includes an example.

Closes #114.


Thomaash:
I just had to resolve conflicts in dist.

PR #125, opened 2019-09-10
fix!(manipulaton-system): don't leak internal objects from edit without drag

### LINKED ISSUE #122: Network - EdgeEdge/EditWithoutDrag return complex objects
The `edgeData` that is passed to the `EditEdge`/`EditWithoutDrag` function is a complex object, rather than a standalone edge (perhaps there's a reason for this, documentation is lacking...) and the callback then fails with "Cannot read property 'connected' of undefined" at e.value (vis-network.esm.min.js:13426).

I suspect this has to do with the fact that `edgeData.from` is an object (the "from" edge object) rather than an ID. The same applies to `edgeData.to`. If I do `edgeData.from = edgeData.from.id` before the callback then the callback succeeds and the network graph is updated correctly.

Happy to do a recreation if you really need it, but I think this explains the problem pretty clearly.

Thomaash:
BREAKING CHANGE: Most people likely used what was returned without checking the documentation and opening an issue as this private object leak has been there for 3 years. Therefore this is very likely to break someone's code even though it just aligns the implementation with the documentation and the older edit with drag, add edge features.

Closes #122.


mojoaxel:
Maybe we should create an example to help with the migration?

Thomaash:
Do we have any channel to communicate these things to the users?

mojoaxel:
> Do we have any channel to communicate these things to the users?

We habe the github release notes. Also an example could be helpfull if the issues start comming in :wink: 

Another way of handling this could be to deprecate the current (wrong) usage, throw a warning to the console and make the breaking change later.

Thomaash:
What if I added a new folder into the examples like this:
```diff
 .
 └── examples
     ├── index.html
     └── network
         ├── basicUsage.html
         ├── exampleUtil.js
---
         ├── other
         │   ├── animationShowcase.html
         │   ├── clusteringByZoom.html
         │   ├── clustering.html
         │   ├── clustersOfclusters.html
         │   ├── configuration.html
         │   ├── cursorChange.html
         │   ├── changingClusteredEdgesNodes.html
         │   ├── chosen.html
         │   ├── manipulationEditEdgeNoDrag.html
         │   ├── manipulation.html
         │   ├── navigation.html
         │   ├── onLoadAnimation.html
         │   ├── performance.html
         │   └── saveAndLoad.html
+        └── upgrading
+            └── 5.4.1 → 6.0.0
+                └── manipulationEditEdgeNoDrag.html
```

mojoaxel:
> What if I added a new folder into the examples like this

Interesting!
Why not! But let's call the folder `5.4.1_to_6.0.0`. I'm a little afraid of unicode names (I had problems before!).

Thomaash:
Yeah, even if it works `5.4.1_to_6.0.0` is a better URL than `5.4.1%20%E2%86%92%206.0.0`.

examples/index.html
```diff
@@ -20,6 +20,9 @@ <h1>Network Examples</h1>
         <h3>basic usage</h3>
         <a class='exampleLink' href="network/basicUsage.html">basic usage</a><br />
 
+        <h3>upgrading</h3>
+        <a class='exampleLink' href="network/upgrading/index.html">upgrading examples</a><br />
```
mojoaxel: I would call "upgrading" better "version migration" or something more clear!

Thomaash: Oukej.

Thomaash: Done.

examples/network/version_migration/5.4.1_to_6.0.0/manipulationEditEdgeNoDrag.html
```diff
@@ -0,0 +1,162 @@
...
+    editEdge: {
+      editWithoutDrag({ from, to }) {
+        const fromId = from.id
+        const toId = to.id
+        …
+      }
+    }
+  }
+}
+    </pre
```
mojoaxel: broken syntax

examples/network/version_migration/5.4.1_to_6.0.0/manipulationEditEdgeNoDrag.html
```diff
@@ -0,0 +1,162 @@
...
+const options = {
+  manipulation: {
+    editEdge: {
+      editWithoutDrag({ from: fromId, to: toId }) {
+        …
+      }
+    }
+  }
+}
+    </pre
```
mojoaxel: s.o.

examples/network/version_migration/5.4.1_to_6.0.0/manipulationEditEdgeNoDrag.html
Thomaash: It's fixed now.

examples/network/version_migration/5.4.1_to_6.0.0/manipulationEditEdgeNoDrag.html
Thomaash: It's fixed now.

PR #140, opened 2019-09-25
chore(examples): generate examples index

Thomaash:
This generates single index file with (potentialy nested) groups and generated screenshots of examples.

The whole thing is written to be reusable, therefore there should be no problem porting this to Graph3D or Timeline. However it get's almost all of it's information from the examples. All examples have to have the canvas in a container with given id (configurable on CLI). Their titles have to follow this structure:
`Group | [Another Group | [Another Group | …]] Example Title`.

There is also some primitive linting that requires the title to match the heading in the body (not enforced in any way, just a warning).

Result: https://jsfiddle.net/thomaash/fn5mzokq/embedded/result (the links would lead somewhere and the images would't be inlined if this was on GitHub Pages instead of JSFiddle).

PS: I tried to fix this real quick but there may still be some issues with example names as they we're a total mess with a lot of them called Basic Usage or whatever they were originally copy pasted from.


mojoaxel:
This could be used to improve https://github.com/visjs/vis-graph3d/issues/2

Thomaash:
Maybe we should create vis-dev-util package for this and potentially other reusable tools created in the future so that we can easily share them across projects.

examples.template.html
```diff
@@ -0,0 +1,101 @@
+<html>
+  <head>
+    <meta charset="utf-8" />
+    <title>Vis Network Examples</title>
+
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+
+    <style>
+      h1 {
```
mojoaxel: I don't think we should any unnecessary headers here. Hopefully at somepoint somebody redesignes this template (we should create an issue for that), but at the moment I would recommend reducing styles in the template to the minimum and let the browsers figure it out how to display it.

examples.template.html
```diff
@@ -0,0 +1,101 @@
...
+      .example-link .example-image > img {
+        position: absolute;
+        top: 0px;
+        left: 0px;
+        width: 200px;
+        height: 200px;
+
+        border: 1px solid #2b7ce9;
+
+        transition: transform 2s ease 0s, z-index 2s linear 0s;
```
mojoaxel: I think 0.5 is quite enough here :wink:

examples.template.html
```diff
@@ -0,0 +1,101 @@
+<html>
```
mojoaxel: I cont think that the root directory is the best place to put all this. Maybe we introduce a "scripts" folder or something!?

mojoaxel:
It might be a good idea to give every network a "seed" so that they look the same every time they get rendered. This blows up the demo-code a little bit but maybe be could inject this option only while rendering!?

generate-example-index.ts
```diff
@@ -0,0 +1,660 @@
...
+    form.append(
+      cheerio("<input>")
+        .attr("type", "hidden")
+        .attr("name", "title")
+        .attr("value", example.titles.join(" | "))
+    );
+
+    return form;
+  }
+  private _generateCodePen(example: Example): Cheerio {
```
mojoaxel: Maybe we also could add support for jsbin? :pray: 
https://jsbin.com/help/api/

examples.template.html
mojoaxel: On mobile the fonts are also WAY to big

Thomaash: The template is only for the index file(s). The examples are kept intact, no harm done. They have optional meta headers but these can be avoided by rewriting the example (unless it has some special needs which is the reason why they exists but that should be rare).

generate-example-index.ts
Thomaash: JS Bin seems to have API only for pro JS Bin account (which I don't have).

Thomaash: Also wouldn't committing the token to a public repo be security risk? Couldn't it be abused?

generate-example-index.ts
```diff
@@ -0,0 +1,660 @@
...
+    return {
+      code: {
+        css,
+        html,
+        js
+      },
+      resources
+    };
+  }
+  private _generateJSFiddle(example: Example): Cheerio {
```
mojoaxel: For me jsfiddle always runs into a timeout.

Thomaash: It's not just you! jsfiddle.net is down: http://isup.me/jsfiddle.net .

examples.template.html
Thomaash: Done.

generate-example-index.ts
mojoaxel: > JS Bin seems to have API only for pro JS Bin account (which I don't have).

I love jsbin and I'm using it for almost a decade I think. 
We could talk to @remy maybe he feels supportive and gives us access for free. I don't think we create any relative traffic...

package.json
```diff
@@ -36,6 +36,7 @@
     "build": "npm run types:network && npm run js:network && npm run images:network",
     "watch": "npm run images:network && rollup --watch --config rollup.config.js",
     "watch-dev": "npm run watch-dev",
+    "generate-examples-index": "node -r ./babel.mocha.js generate-example-index.ts -c mynetwork -w https://visjs.github.io/vis-network -lis",
```
mojoaxel: We should think about calling this from within the CI. This could be done together with #95

Thomaash: Yeah, when #95 is merged I'll add this to it.

Thomaash: Or the other way around if this is merged first.

examples.template.html
mojoaxel: I'm talking about the overview index file.

Thomaash: Why? Does it matter? It's just an index file. Nobody's going to read this anyway.

examples.template.html
Thomaash: Done. It really feels better.

examples.template.html
Thomaash: I din't check it on mobile. That's crazy!

Thomaash: It should be reasonable now.

mojoaxel:
> It might be a good idea to give every network a "seed" so that they look the same every time they get rendered. This blows up the demo-code a little bit but maybe be could inject this option only while rendering!?

@Thomaash What do you think about that?
The option I have in mind is the `randomSeed` in the [layout config](https://visjs.github.io/vis-network/docs/network/layout.html).

examples/network/layout/hierarchicalLayoutDynamicChanges.html
```diff
@@ -0,0 +1,132 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>Vis Network | Layouts | Dynamic Hierarchical Layout Changes</title>
```
mojoaxel: I'm not sure if this PR is the best place to introduce a completely new example, but I don't mind!

generate-example-index.ts
mojoaxel: Ok, let's do this in a later step - if possible...

examples/network/layout/hierarchicalLayoutDynamicChanges.html
Thomaash: This isn't exactly new. It slipped here from #115.

Thomaash:
> > It might be a good idea to give every network a "seed" so that they look the same every time they get rendered. This blows up the demo-code a little bit but maybe be could inject this option only while rendering!?
> 
> @Thomaash What do you think about that?
> The option I have in mind is the `randomSeed` in the [layout config](https://visjs.github.io/vis-network/docs/network/layout.html).

This was a bit tricky but done. I'll have a look at exampleUtil now because it uses Math.random() and screws things up. This should be an easy fix.

Thomaash:
No problem.

PR #142, opened 2019-09-28
chore: update npm badge

mojoaxel:
The npm badge should link to the releases


mojoaxel:
> Shouldn't npm badge link to npm?

:thinking: :bulb: I think you are right!

PR #146, opened 2019-10-01
chore(examples): make index page title configurable

Thomaash:
The title was hardcoded to Vis Network Examples. Now it's reusable across packages.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "build": "npm run types:network && npm run js:network && npm run images:network",
     "watch": "npm run images:network && rollup --watch --config rollup.config.js",
     "watch-dev": "npm run watch-dev",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -36,7 +36,7 @@
     "build": "npm run types:network && npm run js:network && npm run images:network",
     "watch": "npm run images:network && rollup --watch --config rollup.config.js",
     "watch-dev": "npm run watch-dev",
-    "generate-examples-index": "node -r ./babel.mocha.js ./scripts/examples-index/generate.ts -c mynetwork -w https://visjs.github.io/vis-network -S ./scripts/examples-index/screenshot-script.js -lis",
+    "generate-examples-index": "node -r ./babel.mocha.js ./scripts/examples-index/generate.ts -t \"Vis Network Examples\" -c mynetwork -w https://visjs.github.io/vis-network -S ./scripts/examples-index/screenshot-script.js -lis",
```
mojoaxel: We also could use the `title` from the package.json. This could resuilt in something like: `vis-network examples`.

Issue #119, opened 2019-09-06
update issue and PR templates


mojoaxel:
We should update the issue and RP templates on github.
see also:
- https://github.com/visjs/vis-network/community
- https://github.com/visjs/vis-network/tree/master/.github

abhishek-kumar-91:
Still open this issue, I want to work on this issue.

Dxuian:
still open ?

PR #148, opened 2019-10-01
chore(github): update issue template

Thomaash:
There were links pointing to the legacy Vis.js project. Also this no longer contains anything else but Network so mentioning a component is not necessary anymore.


mojoaxel:
This is part of #119

PR #157, opened 2019-10-10
Image arrow heads

### LINKED ISSUE #153: different edge & node types
In context of biological pathway representation we use different edge types & Node types ( reference image attached). This is from Ingenuity Pathway Analysis, some of these are already in vis-network library. However, certain features are missing eg. both bar and arrow, or color filled arrow or blank arrow. I hope these enhancements are generally applied for various graphs. However, I understand all feature requests may not be applicable to all users. In this scenario, is there any example tutorial on how I can write JS snippets that add these new functionality to my networks instead of implementing as part of the complete package ? 

IPA_annalysis

Thomaash:
Adds image arrow heads in similar fashion to image nodes.

Working example: http://jsfiddle.net/thomaash/67xaky4s/.

Todo:
- [x] Implement it.
- [x] Create an example.
- [x] Write docs.

Closes #153.

### DIFF
--- a/docs/network/edges.html
+++ b/docs/network/edges.html
@@ -114,7 +114,7 @@ var options = {
         imageHeight: 32,
         imageWidth: 32,
         scaleFactor: 1,
-        src: "http://placehold.it/32x32/008888",
+        src: "https://visjs.org/images/visjs_logo.png",
         type: "image"
       },
       from: {
--- a/docs/network/edges.html
+++ b/docs/network/edges.html
@@ -101,9 +101,30 @@
 var options = {
   edges:{
     arrows: {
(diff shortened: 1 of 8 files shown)


docs/network/edges.html
```diff
@@ -101,9 +101,30 @@ <h3>Options</h3>
...
+        scaleFactor: 1,
+        src: undefined,
+        type: "arrow"
+      },
+      middle: {
+        enabled: false,
+        imageHeight: 32,
+        imageWidth: 32,
+        scaleFactor: 1,
+        src: "http://placehold.it/32x32/008888",
```
mojoaxel: we should not depend on external links for the docs.

Thomaash: There's an image from our site now.

PR #170, opened 2019-10-21
Fix Mouse scroll not working in FireFox

Tyler-Maclachlan:
A bug we found in our local fork of Visjs Network, especially when other mouse event listeners are added, FireFox now uses 'wheel' instead of 'DOMMouseScroll' - 'DOMMouseScroll' is deprecated: [link](https://developer.mozilla.org/en-US/docs/Web/API/Element/DOMMouseScroll_event)


Thomaash:
Hi @Tyler-Maclachlan,

thanks for your contribution but I have a question. Won't the handler be fired multiple times? Once for each event supported in given browser?

Tyler-Maclachlan:
Hi @Thomaash,

You would think so but no it doesn't. Perhaps due to 'DOMMouseScroll' being deprecated in the browser I'm using.

(It seems 'DOMMouseScroll' fires on mobile sizes, whereas 'wheel' fires on desktop sizes)

Tyler-Maclachlan:
Great will do.

I'm using FireFox 69.0.1 (64-bit) on Fedora Linux and I had to add this to get it to work again. Will update for newer FireFox as well.

Tyler-Maclachlan:
I've been doing some looking and it seems 'wheel' is meant to be the standard.

https://developer.mozilla.org/en-US/docs/Web/API/Document/wheel_event - their example works on all my browsers. I may have forgotten a part of the code I used originally to fix the issue, will update the PR with the changed comments and if I forgot something I'll add it in.

lib/network/modules/InteractionHandler.js
```diff
@@ -511,7 +511,11 @@ class InteractionHandler {
       else if (event.detail) { /* Mozilla case. */
         // In Mozilla, sign of delta is different than in IE.
         // Also, delta is multiple of 3.
+        // For older FireFox versions
         delta = -event.detail / 3;
+      } //Modern FireFox
```
Thomaash: Only Firefox? No other browsers support this?

Also it would look better with a space after `//` and Mozilla doesn't capitalize the second F in Firefox.

lib/network/modules/InteractionHandler.js
```diff
@@ -511,7 +511,11 @@ class InteractionHandler {
...
+      else if (event.deltaY) {
+        delta = -event.deltaY / 3;
```
Thomaash: - Why negative?
- Why divided by 3?

In Firefox I get Y 3 with mode 1 (3 lines) when I scroll, in Chromium it's Y 53 with mode 0 (53 pixels). There should be some handling for this as the way it is now it may produce very different results in different browsers.

Also the modern standard API should be used as the default instead of using it as the last resort solution if everything else fails.

Tyler-Maclachlan:
I'm just gonna change it to @mojoaxel suggestion to just detect which event is supported

Thomaash:
Can't we just use the wheel event only? The other events are for IE8, Opera 17, Firefox 16, Chrome 30. I highly doubt any of these work anyway as we don't include polyfills for them.

lib/network/modules/InteractionHandler.js
```diff
@@ -504,15 +504,22 @@ class InteractionHandler {
...
-        // In Mozilla, sign of delta is different than in IE.
-        // Also, delta is multiple of 3.
-        delta = -event.detail / 3;
-      }
+      /*
+        No need for if statement as new wheel listener function
+        normalizes the event to the 'wheel' api
+      */
+      let delta = event.deltaY;
+      // if (event.deltaY) { /* Modern browsers */
```
mojoaxel: this comments should be removed before mergin!

lib/network/modules/InteractionHandler.js
```diff
@@ -521,11 +528,15 @@ class InteractionHandler {
...
-        let zoom = delta * (this.options.zoomSpeed / 10);
-        if (delta < 0) {
-          zoom = zoom / (1 - zoom);
-        }
-        scale *= (1 + zoom);
+        let direction = delta < 0 ? 1 : -1;
+        // times is faster than divide
+        // not using delta creates a more consistent experience across browsers
+        let zoom = 1 + direction * (this.options.zoomSpeed * 0.1);
+        // let zoom = delta * (this.options.zoomSpeed / 10);
```
mojoaxel: remove comments

lib/network/modules/InteractionHandler.js
Tyler-Maclachlan: Removed

lib/network/modules/InteractionHandler.js
Tyler-Maclachlan: Removed

Tyler-Maclachlan:
@Thomaash 
We can't only use 'wheel' as just 2 versions of Firefox ago uses 'DOMMouseScroll'

Thomaash:
@Tyler-Maclachlan I have no idea where you got this but just to be sure I deleted all the other events from the library and kept only wheel. Then I downgraded to Firefox 52 (the oldest we include polyfills for) and tested it. Zooms without an issue. According to https://caniuse.com/#search=wheel this should go all the way back to Firefox 17. I didn't test that though as we don't include polyfills for that. It probably doesn't work anyway.

Since the only browser that has ever supported DOMMouseScroll seems to be Firefox and it has support for wheel in all of the versions we support I see absolutely no reason to keep DOMMouseScroll.

Tyler-Maclachlan:
I have removed the other listeners :+1:

lib/network/modules/Canvas.js
```diff
@@ -249,8 +250,7 @@ class Canvas {
     this.hammer.on('pinch',     (event) => {this.body.eventListeners.onPinch(event)});
 
     // TODO: neatly cleanup these handlers when re-creating the Canvas, IF these are done with hammer, event.stopPropagation will not work?
-    this.frame.canvas.addEventListener('mousewheel',     (event) => {this.body.eventListeners.onMouseWheel(event)});
-    this.frame.canvas.addEventListener('DOMMouseScroll', (event) => {this.body.eventListeners.onMouseWheel(event)});
```
Thomaash: Why? I see absolutely no reason to do this. Just delete DOMMouseScroll and replace mousewheel by wheel.

lib/network/modules/InteractionHandler.js
```diff
@@ -504,15 +504,7 @@ class InteractionHandler {
...
-      let delta = 0;
-      if (event.wheelDelta) { /* IE/Opera. */
-        delta = event.wheelDelta / 120;
-      }
-      else if (event.detail) { /* Mozilla case. */
-        // In Mozilla, sign of delta is different than in IE.
-        // Also, delta is multiple of 3.
-        delta = -event.detail / 3;
-      }
+      let delta = event.deltaY;
```
Thomaash: The keyword let screams that this variable will be reassigned but it never is. Always use const unless you reassign the variable later.

lib/network/modules/InteractionHandler.js
```diff
@@ -521,11 +513,11 @@ class InteractionHandler {
 
         // calculate the new scale
         let scale = this.body.view.scale;
-        let zoom = delta * (this.options.zoomSpeed / 10);
-        if (delta < 0) {
-          zoom = zoom / (1 - zoom);
-        }
-        scale *= (1 + zoom);
+        let direction = delta < 0 ? 1 : -1;
```
Thomaash: Again if you don't change it don't use let, use const instead.

PS: I know that's the way it was but the old code is a mess. Don't take inspiration from that.

lib/network/modules/InteractionHandler.js
```diff
@@ -521,11 +513,11 @@ class InteractionHandler {
...
+        // times is faster than divide
+        // not using delta creates a more consistent experience across browsers
+        let zoom = 1 + direction * (this.options.zoomSpeed * 0.1);
+        scale *= zoom;
```
Thomaash: Why the variable? Seems to be used once at the next line and then thrown away.

lib/network/modules/MouseWheelListener.js
```diff
@@ -0,0 +1,34 @@
...
+let prefix = "",
+_addEventListener,
+support;
+
+if (window.addEventListener) {
+    _addEventListener = 'addEventListener';
+} else {
+    _addEventListener = 'attachEvent';
+    prefix = "on";
+}
```
Thomaash: This is IE8 stuff. The oldest browser we support is IE11. Not mentioning the fact that the rest of the codebase doesn't use attachEvent anyway.

This file as whole is useless.

Tyler-Maclachlan:
I messed up my fork so I had to re-fork vis, I created a new PR, #208 for the new fork.

I get what you meant now, and I think the code is a lot cleaner now.

PR #171, opened 2019-10-21
feat(hierarchical-layout): add avoid overlap

### LINKED ISSUE #168: Hierarchical layout to respect node size
I am working with a hierarchical layout.
My nodes carry labels of  varying length.
AFAIK, the standard physics solver `hierarchicalRepulsion` computes a repulsion only based on the node centre. It does not take the node size (with label) into account.

Would it be possible to implement such a solver? 

Bildschirmfoto 2019-10-20 um 13 37 42

For hierachical layouts, it should be even be possible to position all nodes **without physics simulation**: All one needs is a margin between neighboring nodes (i use `direction: "UD",				sortMethod: "directed"`).

image

Thomaash:
The same thing as in Barnes Hut.

Includes an example and docs.

Live example: https://jsfiddle.net/thomaash/cmaojfg2/.

Closes #168.

### DIFF
--- a/docs/network/physics.html
+++ b/docs/network/physics.html
@@ -117,7 +117,8 @@ var options = {
       springLength: 100,
       springConstant: 0.01,
       nodeDistance: 120,
-      damping: 0.09
+      damping: 0.09,
+      avoidOverlap: 0
     },
     maxVelocity: 50,
     minVelocity: 0.1,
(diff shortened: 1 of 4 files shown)


flaushi:
Ah wow! That's so great! Thanks so much! When will the next release including this one be available?

PR #185, opened 2019-10-27
chore(examples): fix broken script paths

Thomaash:
Some paths in the examples pointed to nonexistent files. This fixes that.


Qix-:
Doesn't seem to be fixed on the site.

Thomaash:
This is fixed, this however affected only two examples.

Qix-:
image

Thomaash:
I know but I'm 99.99 % sure this is a different issue.

PR #187, opened 2019-10-28
test(e2e): add Cypress

Thomaash:
This includes three very simple tests using the manipulation GUI (click the canvas and check the coordinates; place a node and check it's placed; place two nodes, connect them with an edge and check the edge is there). These should ensure that the library is not completely broken though they test only a very tiny part of the library.

PS: Just in case there is some confusion this doesn't replace #35. We still should figure out how to make that work.


Thomaash:
You have to build first: `npm run build && npm ci && npm test`.

mojoaxel:
> npm run build && npm ci && npm test

If `test` depends on `build`...shoulden't we have a dependency in the npm scripts!?

PR #223, opened 2019-11-10
feat(build): add index files to bundles

Thomaash:
This allows more visually appealing imports when used with bundlers:
```typescript
// Old, but still works:
import * as vis from "vis-network/standalone/esm/vis-network";

// New, achieving the same result (ESM):
import * as vis from "vis-network/standalone";

// ESM explicitly:
import * as vis from "vis-network/standalone/esm";

// UMD explicitly:
import * as vis from "vis-network/standalone/umd";
```

This partially works in HTML but the old way is preferred (less requests,
library name in dev tools and better compatibility):
```html
<!-- Old, still works. Recommended. -->
<script
  type="text/javascript"
  src="https://unpkg.com/vis-network/standalone/umd/vis-network.min.js"
></script>

<!-- New, works but has issues. Not recommended. -->
<script
  type="module"
  src="https://unpkg.com/vis-network/standalone/esm"
></script>

<!-- New, doesn't work. -->
<script
  type="text/javascript"
  src="https://unpkg.com/vis-network/standalone/umd"
></script>
```

### DIFF
--- a/rollup.build.js
+++ b/rollup.build.js
@@ -80,6 +80,46 @@ export default [].concat.apply(
                     src: `./dev-lib/bundle-${variant}.d.ts`,
                     dest: ".",
                     rename: umdFileWithoutExt + ".d.ts"
+                  },
+                  {
+                    src: "./dev-lib/bundle-esm-index.d.ts",
+                    dest: ".",
+                    rename: `${variant}/index.d.ts`
+                  },
+                  {
+                    src: "./dev-lib/bundle-esm-index.d.ts",
+                    dest: ".",
+                    rename: `${variant}/index.js`
+                  },
+                  {
+                    src: "./dev-lib/bundle-vis-network-index.d.ts",
+                    dest: ".",
+                    rename: `${variant}/esm/index.d.ts`
+                  },
+                  {
+                    src: "./dev-lib/bundle-vis-network-index.d.ts",
+                    dest: ".",
+                    rename: `${variant}/esm/index.js`
+                  },
+                  {
+                    src: "./dev-lib/bundle-vis-network-index.d.ts",
+                    dest: ".",
+                    rename: `${variant}/umd/index.d.ts`
+                  },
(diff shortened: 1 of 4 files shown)


rollup.build.js
```diff
@@ -80,6 +80,46 @@ export default [].concat.apply(
                     src: `./dev-lib/bundle-${variant}.d.ts`,
                     dest: ".",
                     rename: umdFileWithoutExt + ".d.ts"
+                  },
+                  {
+                    src: "./dev-lib/bundle-esm-index.d.ts",
+                    dest: ".",
+                    rename: `${variant}/index.d.ts`
```
yotamberk: Just a suggestion, but maybe you can right this in a cleaner way. Something like:
```javascript
const jsNames = [
'index.js',
'esm/index.d.ts',
'esm/index.js',
'umd/index.d.ts',
'umd/index.js',
]

const cssNames = [
'index.d.ts',
'index.js'
]

const jsBuilds = jsNames.map((jsName) => ({
  src: "./dev-lib/bundle-vis-network-index.d.ts",
  dest: ".",
  rename: `${variant}/${jsName}`
}))

const cssBuilds = cssNames.map((cssName) => ({
  src: "./dev-lib/bundle-vis-network-index.d.ts",
  dest: ".",
  rename: `styles/${cssName}`
}))
```

and then add them to the builds?

Thomaash: I though about such solution when I was writing this but I'm not sure it's clearer. The way it is now it basically says copy A to B, A to C, D to E, D to F, etc. Pretty straightforward and easy to understand albeit verbose. With your solution there would be a bunch of arrays and mapping over them and it would actually take about the same number of lines of code (your code won't work because index.js is different from esm/index.js, it would require separate array and map) to achieve the same in quite cryptic and hard to follow way.

yotamberk: Ok. Sounds fair to me.

PR #230, opened 2019-11-12
feat(edges): make self reference position configurable

### LINKED ISSUE #200: Self reference circle position option
I would like to request features / properties related to the selfReference. Property to say that selfReference edge is at the top / bottom / right / left (meaning the edge circle position is centered based on node according to provided position). 

I am attaching images to represent my requested option:

So the following image is something i would expect from requested options. You can see that  self ref is at the center on top.

Capture

piratuks:
So it can be tested with the following setup:

    var nodes = new vis.DataSet([
        {id: 1, shape:"box", label: 'Node 1', title: 'I have a popup!', widthConstraint:{minimum:270}, heightConstraint:{minimum:200}, color: {background: "transparent"}},
        {id: 2, shape:"box", label: 'Node 2', title: 'I have a popup!', widthConstraint:{minimum:270}, heightConstraint:{minimum:200}, color: {background: "transparent"}},
        {id: 3, shape:"box", label: 'Node 3', title: 'I have a popup!', widthConstraint:{minimum:270}, heightConstraint:{minimum:200}, color: {background: "transparent"}},
        {id: 4, shape:"box", label: 'Node 4', title: 'I have a popup!', widthConstraint:{minimum:270}, heightConstraint:{minimum:200}, color: {background: "transparent"}}
    ]);

    var edges = new vis.DataSet([
        {label:"Test", from: 1, to: 1, arrows:'to, middle, from'},
        {label:"Test", from: 2, to: 2, arrows:'to, middle, from'},
        {label:"Test", from: 3, to: 3, arrows:'to, middle, from'},
        {label:"Test", from: 4, to: 4, arrows:'to, middle from'},
    ]);
    
OR

    var edges = new vis.DataSet([
        {label:"Test", from: 1, to: 1, arrows:'to, middle, from', selfReferenceSize: 20, selfReference: {size: 40, side:"left"}},
        {label:"Test", from: 2, to: 2, arrows:'to, middle, from', selfReferenceSize: 20},
        {label:"Test", from: 3, to: 3, arrows:'to, middle, from', selfReference: {size: 20, side:"right"}},
        {label:"Test", from: 4, to: 4, arrows:'to, middle from', selfReference: {side:"left"}},
    ]);

OR

    var data = {
        nodes: nodes,
        edges: edges,
        options:{
            edges:{
                selfReferenceSize:60,
                selfReference: {side:"right"}
            }
        }
    };

fixes #200


lib/network/modules/components/Edge.js
```diff
@@ -677,16 +677,46 @@ class Edge {
...
-        if (node1.shape.width > node1.shape.height) {
+        var radius = (typeof this.options.selfReferenceSize === "number")? this.options.selfReferenceSize : this.options.selfReferenceSize.size;
+        
+        // get circle coordinates
+        if(this.options.selfReferenceSize.side !== null && typeof this.options.selfReferenceSize !== "number"){
+          if(this.options.selfReferenceSize.side === "top"){
+            //top center
+            x = node1.x;
+            y = node1.y - node1.shape.height * 0.5;            
+            point = this._pointOnCircle(x, y, radius, 0.125);
```
Thomaash: Is it just me or is this the same as...

lib/network/modules/components/Edge.js
```diff
@@ -677,16 +677,46 @@ class Edge {
...
+          }else if(this.options.selfReferenceSize.side === "right"){
+            //right center
+            x = node1.x + node1.shape.width * 0.5;
+            y = node1.y;             
+            point = this._pointOnCircle(x, y, radius, 0.9);
+          }else{
+            //top center
+            x = node1.x;
+            y = node1.y - node1.shape.height * 0.5;
+            point = this._pointOnCircle(x, y, radius, 0.125);
```
Thomaash: ...this and...

lib/network/modules/components/Edge.js
```diff
@@ -677,16 +677,46 @@ class Edge {
...
+        }else if (node1.shape.width > node1.shape.height) {
           x = node1.x + node1.shape.width * 0.5;
           y = node1.y - radius;
-        }
-        else {
+          point = this._pointOnCircle(x, y, radius, 0.125);
+        } else {
           x = node1.x + radius;
           y = node1.y - node1.shape.height * 0.5;
+          point = this._pointOnCircle(x, y, radius, 0.125);
```
Thomaash: ...this?

test/edges/cubic-bezier-edge.test.ts
```diff
@@ -146,7 +146,9 @@ describe('CubicBezierEdge', function(): void {
           smooth: {
             roundness: 2,
           },
-          selfReferenceSize: 42,
+          selfReferenceSize: {
+            size: 42
+          },
```
Thomaash: Why? Is this a breaking change? Number wouldn't work anymore?

piratuks: No it is not, that can be as it was, you are right.

lib/network/modules/components/Edge.js
piratuks: What do you mean by the same, you mean the line which is calling _pointOnCircle if that is the case then yea it is the same but params are diff. So i guess i could just set new variable which holds diff percentage for each case and then after case sentences call it.

Thomaash: Line 685 === line 705.
Line 686 === line 706.
Line 687 === line 707.
Line 688 === line 708.

lib/network/modules/components/Edge.js
Thomaash: This one differs. I missed the radius on line 715.

lib/network/modules/components/Edge.js
piratuks: Yes, my thinking was this: i had these 4 options for position so i check these and if property is set to something weird which is not within property range [top,right,bottom,left] then i set top as default.

piratuks:
>     * [ ]  Could you make it so that the angle can be chosen freely? For example `Math.PI / 2` for right, `0` for top but also `Math.PI / 4` for topright.


What do you mean by this ?

Thomaash:
> > ```
> > * [ ]  Could you make it so that the angle can be chosen freely? For example `Math.PI / 2` for right, `0` for top but also `Math.PI / 4` for topright.
> > ```
> 
> What do you mean by this ?

For example
```javascript
selfReference: {
  size: 30,
  angle: Math.PI / 4
}
```
instead of
```javascript
selfReference: {
  size: 30,
  side: "right"
}
```

It would allow the circle to be placed anywhere around the node not just at few predefined angles.

piratuks:
@Thomaash  I think i made changes based on your request. Created diff property as we discussed  With deprecated selfReferenceSize for backwards compatibility. However the only change i have not touched is that angle thing. Could expand on the logic, because i am not really understand how you gonna calculate x and y for the circle if you provide angle. Because right now if i say top then mean x is the same as the parent node and y is the top border of parent node. So now if you provide Math.PI / 4 how is that gonna work?

Thomaash:
@piratuks I don't know if this is the best solution but you can use `Node.distanceToBorder()` to get the distance from the center of the node to the border of the node at given angle. This way you'll get the polar coordinates of the point you want relative to the center of the node. The rest is elementary school math.

lib/network/Network.js
```diff
@@ -344,6 +344,16 @@ Network.prototype.setData = function (data) {
...
 
+  if(data && data.options && data.options.edges && data.options.edges.hasOwnProperty("selfReferenceSize")){
+    console.log('The selfReferenceSize property has been deprecated. Please use selfReference property instead. The selfReference can be set like thise selfReference:{size:30, angle:null}');
+    if(!data.options.edges.hasOwnProperty("selfReference")){
+      data.options.edges.selfReference = {
+          size: data.options.edges.selfReferenceSize,
+          angle: null
+      }
+    }
+  }
```
Thomaash: This belongs into `Edge.parseOptions()`.

lib/network/modules/EdgesHandler.js
```diff
@@ -94,7 +94,11 @@ class EdgesHandler {
         }
       },
       selectionWidth: 1.5,
-      selfReferenceSize:20,
+      selfReferenceSize: 20,
+      selfReference: { 
+        size: 20,
+        angle: null,
```
Thomaash: Why null? Can't it be undefined like everything else?

types/network/Network.d.ts
```diff
@@ -968,6 +968,11 @@ export interface EdgeOptions {
 
   selfReferenceSize?: number;
 
+  selfReference?: {
+    size: number,
```
Thomaash: Why is this mandatory?

Thomaash:
Also trying to build your repo and run with (modified Basic Usage):
```javascript
// create an array with nodes
var nodes = new vis.DataSet([
  { id: 1, label: "Node 1" },
  { id: 2, label: "Node 2" },
  { id: 3, label: "Node 3" },
  { id: 4, label: "Node 4" },
  { id: 5, label: "Node 5" }
]);

// create an array with edges
var edges = new vis.DataSet([
  { from: 1, to: 3 },
  { from: 1, to: 2 },
  { from: 2, to: 4 },
  { from: 2, to: 5 },
  { from: 3, to: 3 }
]);

// create a network
var data = {
  nodes: nodes,
  edges: edges
};
var options = {};

(async () => {
  let angle = -10;

  for (;;) {
    await new Promise(resolve => setTimeout(resolve, 1000));

    ++angle;
    console.log("set self reference angle to", angle / 10);
    network.setOptions({
      edges: {
        selfReference: {
          angle: angle / 10
        }
      }
    });
  }
})();
```
It doesn't do anything. Not even any errors. It's like if the option was never even set.

piratuks:
@Thomaash thx for quick feedback. i am gonna check your comments. Was off for a bit so that is why no progress on this PR. Now i am back wanna finish this one.

piratuks:
@Thomaash I updated PR. Added changes based on your recent feedback, some comments for most important parts and checked your example. Please try again and give me your thoughts.

docs/network/edges.html
```diff
@@ -894,7 +898,31 @@ <h3>Options</h3>
...
+            <td>This is the radius of circle.
+            </td>
+        </tr>
+        <tr parent="selfReference" class="hidden">
+            <td class="indent">angle</td>
+            <td>Number</td>
+            <td><code>undefined</code></td>
+            <td> This is the angle (Provided in radians) to indicate position for the circle. If position is not provided then it will 
+                be automatically determined.
+            </td>
```
Thomaash: Could you add this to the Korean docs, please?

PS: If you don't speak Korean then instead of translating it simply replace the English text with `TODO`. Just so that we know there's something in need of translating. Thanks.

lib/network/modules/components/Edge.js
```diff
@@ -677,16 +686,73 @@ class Edge {
         // Ignore the orientations.
         this.labelModule.pointToSelf = true;
         var x, y;
-        var radius = this.options.selfReferenceSize;
-        if (node1.shape.width > node1.shape.height) {
+
+        var radius = this.options.selfReference.size;
+        var percentage = 0.125;
+
```
Thomaash: If you change the `percentage` to angle you'll be able to simply assign the self reference angle to it and have the edge label positioned away from the node no matter what angle was chosen which, at least in my opinion, would be much better. Also faster and simpler code.

As a bonus this would work correctly for all angles including negative or angles in thousands etc.

PS: Quick searched revealed that `Edge._pointOnCircle` is used only from this method so you can safely remove the percentage to angle conversion from it.

lib/network/modules/components/Edge.js
```diff
@@ -677,16 +686,73 @@ class Edge {
...
+          x = node1.x;
+          y = node1.y;
+          
+          if(ctx){
+            //calculating opposite and adjacent
+            //distaneToBorder becomes Hypotenuse. 
+            //Formulas sin(a) = Opposite / Hypotenuse and cos(a) = Adjacent / Hypotenuse
+            const toBorderDist = node1.distanceToBorder(ctx, this.options.selfReference.angle);
+            const sin = Math.sin(this.options.selfReference.angle)*toBorderDist;
+            const cos = Math.cos(this.options.selfReference.angle)*toBorderDist;
```
Thomaash: Isn't `sin` the x distance from the center of the node? Wouldn't something like `xFromNodeCenter` be more descriptive?
dtto for `cos` on y axis.

lib/network/modules/components/Edge.js
```diff
@@ -677,16 +686,73 @@ class Edge {
...
+              if(sin > 0){
+                y -= Math.abs(sin);
+              }else{
+                y += Math.abs(sin);
+              }
+              if(cos > 0){
+                x += Math.abs(cos);
+              }else{
+                x -= Math.abs(cos);
+              }
```
Thomaash: Isn't this equal to the following?
```typescript
x += cos;
y -= sin;
```

lib/network/modules/components/edges/util/edge-base.ts
```diff
@@ -322,13 +337,59 @@ export abstract class EdgeBase<Via = undefined> implements EdgeType {
...
+        }
+      }
+    } else if (node.shape.width > node.shape.height) {
       x = node.x + node.shape.width * 0.5;
       y = node.y - radius;
     } else {
       x = node.x + radius;
       y = node.y - node.shape.height * 0.5;
     }
+
```
Thomaash: I didn't compare this character by character but isn't this the same logic as in Edge.js? If so could you create a reusable function for this rather than hardcoding it in two places, please?

piratuks:
@Thomaash  thanks for comments. All makes sense. However i am not quite sure what to do with from, to, middle arrows. Any help / comment on this ?

piratuks:
@Thomaash  I think i figured the arrow issue aswell. Also added changes from last feedback. Please check again.

Thomaash:
👏 It works. There are a few details that need polishing and it's ready to go.

lib/network/modules/EdgesHandler.js
```diff
@@ -94,7 +94,11 @@ class EdgesHandler {
         }
       },
       selectionWidth: 1.5,
-      selfReferenceSize:20,
+      selfReferenceSize: 20,
+      selfReference: { 
+        size: 20,
+        angle: undefined,
+      },
```
Thomaash: Maybe you could set default angle to the top right corner to simplify the logic later on (no need to check if angle is set because it always would).

lib/network/modules/components/Edge.js
```diff
@@ -249,6 +250,14 @@ class Edge {
     if (allowDeletion === true && newOptions.font === null) {
       parentOptions.font = util.bridgeObject(globalOptions.font); // set the object back to the global options
     }
+
+    if(newOptions.hasOwnProperty("selfReferenceSize")){
+      console.log('The selfReferenceSize property has been deprecated. Please use selfReference property instead. The selfReference can be set like thise selfReference:{size:30, angle:null}');
```
Thomaash: Don't tell people to set angle to null when null is reported as invalid value.

lib/network/modules/components/Edge.js
```diff
@@ -676,17 +685,29 @@ class Edge {
...
-        if (node1.shape.width > node1.shape.height) {
-          x = node1.x + node1.shape.width * 0.5;
-          y = node1.y - radius;
-        }
-        else {
-          x = node1.x + radius;
-          y = node1.y - node1.shape.height * 0.5;
-        }
-        point = this._pointOnCircle(x, y, radius, 0.125);
+        var angle = 0.125 * 2 * Math.PI;
```
Thomaash:
```javascript
let angle =
```

lib/network/modules/components/Edge.js
```diff
@@ -676,17 +685,29 @@ class Edge {
...
+        if (
+          typeof this.options.selfReference.angle !== "undefined" &&
+          this.options.selfReference.angle !== null
```
Thomaash: This is not important but `null == undefined` so `this.options.selfReference.angle != null` would achieve the same result.

lib/network/modules/components/Edge.js
```diff
@@ -676,17 +685,29 @@ class Edge {
...
+        ) {
+          angle = this.options.selfReference.angle;
+        } 
+
+        // get circle coordinates
+        var coordinates = ComponentUtil.getSelfRefCoordinates(
```
Thomaash:
```javascript
const coordinates =
```

lib/network/modules/components/edges/util/edge-base.ts
```diff
@@ -106,6 +107,21 @@ export abstract class EdgeBase<Via = undefined> implements EdgeType {
...
+      if (
+        !Object.prototype.hasOwnProperty.call(options, "selfReference") ||
+        !Object.prototype.hasOwnProperty.call(options.selfReference, "size")
+      ) {
+        this.options.selfReference = {
+          size: options.selfReferenceSize
+        };
+      }
+    }
+
```
Thomaash: This is redundant. It was already done in `Edge.parseOptions()`.

lib/network/modules/components/edges/util/edge-base.ts
```diff
@@ -746,28 +760,61 @@ export abstract class EdgeBase<Via = undefined> implements EdgeType {
...
+          high = this.options.selfReference.angle;
+        }
+
         const pointT = this._findBorderPositionCircle(this.from, ctx, {
           x,
           y,
-          low: 0.25,
-          high: 0.6,
+          low: low,
+          high: high,
```
Thomaash: For consistency sake I would prefer:
```javascript
x,
y,
low,
high,
```

lib/network/modules/components/edges/util/edge-base.ts
```diff
@@ -746,28 +760,61 @@ export abstract class EdgeBase<Via = undefined> implements EdgeType {
...
+          high = this.options.selfReference.angle;
+        }
+
         const pointT = this._findBorderPositionCircle(this.from, ctx, {
           x,
           y,
-          low: 0.6,
-          high: 1.0,
+          low: low,
+          high: high,
```
Thomaash: dtto

lib/network/modules/components/edges/util/edge-base.ts
```diff
@@ -746,28 +760,61 @@ export abstract class EdgeBase<Via = undefined> implements EdgeType {
...
         });
         angle = pointT.t * -2 * Math.PI + 1.5 * Math.PI - 1.1 * Math.PI;
         arrowPoint = pointT;
       } else {
         arrowPoint = this._pointOnCircle(x, y, radius, 0.175);
         angle = 3.9269908169872414; // === 0.175 * -2 * Math.PI + 1.5 * Math.PI + 0.1 * Math.PI;
+        //changing low and high for the arrows to avoid overlapping with the parent shape
+        if (
+          typeof this.options.selfReference.angle !== "undefined" &&
+          this.options.selfReference.angle !== null
```
Thomaash: dtto

lib/network/modules/components/shared/ComponentUtil.js
```diff
@@ -114,6 +114,64 @@ class ComponentUtil {
...
+   *
+   * @param {Object} ctx
+   * @param {number} angle 
+   * @param {number} radius 
+   * @return {Object} node
+   * @returns {Object} x and y coordinates
+   */
+  static getSelfRefCoordinates(ctx, angle, radius, node){
+    var x = node.x,
+    y = node.y;
```
Thomaash:
```javascript
let x =
let y =
```

Thomaash:
Oh, you're changing it while I'm reviewing it :smile:.

piratuks:
Just added the example html. Though it would be useful (its your example). Will do the changes then

lib/network/modules/components/edges/util/edge-base.ts
piratuks: @Thomaash  Are you sure? 
I modified it to be as the following:

if (Object.prototype.hasOwnProperty.call(options, "selfReferenceSize")) {
      if (
        !Object.prototype.hasOwnProperty.call(options, "selfReference") ||
        !Object.prototype.hasOwnProperty.call(options.selfReference, "size")
      ) {
        this.options.selfReference = {
          size: options.selfReferenceSize,
          angle: null
        };
      }
    }


Because if we remove it then unit tests need to be changed from selfReferenceSize to selfReference: {size, angle}, because passed option does not contain selfReference. Unless i missed something.

Thomaash: In production this should never get raw options only the options that were processed by `Edge.parseOptions()` (any other behavior is a bug). If the tests test this directly then they should be updated to test the new API not the deprecated one.

piratuks:
@Thomaash So i changed test scripts to test new version of self ref. Added changes from last feedback. However there is just one thing i did not understood. When running unit tests it shows under console that self ref size deprecate msg even though i provide under each test the new property. So that was weird. Could you check the changes and let me know regarding if you see anything else?

Thomaash:
Just add `console.trace()` after the log message to see the callstack it came through.

docs-kr/network/edges.html
```diff
@@ -204,6 +204,10 @@ <h3>옵션</h3>
     },
     selectionWidth: 1,
     selfReferenceSize:20,
+    selfReference:{
+        size: 20,
+        angle: Math.PI / 2
```
Thomaash: This was originally π/4 and is π/4 in the code.

docs-kr/network/edges.html
```diff
@@ -886,6 +890,27 @@ <h3>옵션</h3>
...
+            <td><span parent="arrows" class="right-caret"></span> selfReference</td>
+            <td>Object</td>
+            <td><code>Object</code></td>
+            <td>TODO
+            </td>
+        </tr>
+        <tr parent="selfReference" class="hidden">
+            <td class="indent">size</td>
+            <td>Number</td>
+            <td><code>false</code></td>
```
Thomaash: The default value for datatype number cannot be false. In the code it was and is 20.

docs-kr/network/edges.html
```diff
@@ -886,6 +890,27 @@ <h3>옵션</h3>
...
+            <td>TODO
+            </td>
+        </tr>
+        <tr parent="selfReference" class="hidden">
+            <td class="indent">angle</td>
+            <td>Number</td>
+            <td><code>undefined</code></td>
```
Thomaash: This was originally π/4 and is π/4 in the code.

docs/network/edges.html
```diff
@@ -203,7 +203,11 @@ <h3>Options</h3>
       }
     },
     selectionWidth: 1,
-    selfReferenceSize:20,
+    selfReferenceSize: 20,
+    selfReference:{
+        size: 20,
+        angle: Math.PI / 2
```
Thomaash: This was originally π/4 and is π/4 in the code.

docs/network/edges.html
```diff
@@ -895,7 +899,31 @@ <h3>Options</h3>
...
+            <td><span parent="arrows" class="right-caret"></span> selfReference</td>
+            <td>Object</td>
+            <td><code>Object</code></td>
+            <td>When the to and from nodes are the same, a circle is drawn. This provides radius of circle and position.
+            </td>
+        </tr>
+        <tr parent="selfReference" class="hidden">
+            <td class="indent">size</td>
+            <td>Number</td>
+            <td><code>false</code></td>
```
Thomaash: The default value for datatype number cannot be false. In the code it was and is 20.

docs/network/edges.html
```diff
@@ -895,7 +899,31 @@ <h3>Options</h3>
...
+            <td>This is the radius of circle.
+            </td>
+        </tr>
+        <tr parent="selfReference" class="hidden">
+            <td class="indent">angle</td>
+            <td>Number</td>
+            <td><code>undefined</code></td>
```
Thomaash: This was originally π/4 and is π/4 in the code.

docs/network/edges.html
```diff
@@ -895,7 +899,31 @@ <h3>Options</h3>
...
+            <td> This is the angle (Provided in radians) to indicate position for the circle. If position is not provided then it will 
+                be automatically determined.
```
Thomaash: What do you mean by automatically determined? Shouldn't it just be in the top right corner?

piratuks:
@Thomaash  console.trace helped to find where the msg comes from. Should be all good now.
Let me know if you see any other change.

piratuks:
@Thomaash  I updated comments. Also wanted to say thanks for your patient, guidance and cooperation :)

Thomaash:
Regarding the snapshot tests I was afraid they would be very fragile. Could you open a PR for this and share some info about your environment? I would totally understand if you don't want to go down this rabbit hole but it would help a lot.

Thanks either way.

piratuks:
> 
> 
> Regarding the snapshot tests I was afraid they would be very fragile. Could you open a PR for this and share some info about your environment? I would totally understand if you don't want to go down this rabbit hole but it would help a lot.
> 
> Thanks either way.

@Thomaash  Let me just confirm because i am not quite sure if i understood. You want for me to open PR with the values which was expected or a issue?

Thomaash:
An issue.

PR #270, opened 2019-11-26
fix(hierarchical): ignore invisible nodes

### LINKED ISSUE #264: Hierarchical layout creates massive gaps when there are nodes in clusters
Hi guys,

So I have a graph of around ~930 nodes

916 of those nodes are connected to one node and therefore are clustered into that node.

The problem is the layout still creates space for them and thus creates large gaps where those nodes would be if they weren't clustered.

This creates a scenario where I have a node that is clustered and I have to scroll down forever to find the 13 nodes that aren't clustered because there is still space given for the 916 nodes that are in the cluster.

Is there any way for the layout to not give space for the nodes in the cluster or for the layout to order the nodes that aren't in the cluster above the nodes that are in the cluster?

I am clustering all the nodes before I enable the hierarchical layout and all the nodes have predefined levels.

My layout config:
`hierarchical: {
    enabled: false,
    direction: 'RL',
    sortMethod: 'directed',
    nodeSpacing: 200,
    treeSpacing: 200,
    levelSeparation: 250
}`

The layout `enabled` gets changed after clustering is done.

large_network_gap

Attached screen shot of what I mean, the nodes at the top are where the cluster is and alll the way at the bottom is a lonely node that is on the same level as the nodes in the cluster but gets pushed way down because there is space allocated for the nodes that are in the cluster.

Thomaash:
The layout originally made space for invisible nodes (for example after clustering). Now it takes into account only the nodes that are actually rendered solving the problems with sometimes huge empty spaces in hierarchies with clusters.

Also the Hierarchical Layout Methods example was modified to show clustering. It was actually mostly rewritten with new design, various code style improvements and more detailed info added atop of clustering.

PS: Merge #268 into this branch to make the errors go away.

- Closes #264.

### DIFF
--- a/examples/network/layout/hierarchicalLayoutMethods.html
+++ b/examples/network/layout/hierarchicalLayoutMethods.html
@@ -1,126 +1,298 @@
-<!doctype html>
+<!DOCTYPE html>
 <html>
-<head>
-  <title>Vis Network | Layouts | Hierarchical Layout Difference</title>
+  <head>
+    <title>Vis Network | Layouts | Hierarchical Layout Methods</title>
 
-  <style type="text/css">
-    body {
-      font: 10pt sans;
-    }
-    #mynetwork {
-      width: 800px;
-      height: 500px;
-      border: 1px solid lightgray;
-    }
-  </style>
-
-  <script type="text/javascript" src="../../../standalone/umd/vis-network.min.js"></script>
-
-  <script type="text/javascript">
-    var network = null;
-
-    function destroy() {
-      if (network !== null) {
-        network.destroy();
-        network = null;
+    <style type="text/css">
+      html,
+      body,
+      #mynetwork {
+        margin: 0px;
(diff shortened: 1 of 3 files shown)


EquiAvia:
Very keen for this to be merged in, will solve a number of problems I have been experiencing.

👍

Thomaash:
Hi @EquiAvia,

are there any other issues this resolves?

EquiAvia:
> Hi @EquiAvia,
> 
> are there any other issues this resolves?

Not from a VisJs perspective, but it resolves a number of user experience challenges I have had with the large extra spaces allocated to the hidden nodes in our app.  Thank you for keeping VisJs alive.

Thomaash:
I'm glad to hear that my work here helps.

Tyler-Maclachlan:
Thanks @Thomaash I will definitly be trying this at work on Monday, but the examples look promising :)

PR #286, opened 2019-12-02
test(e2e): add directed hierarchical layout test

Thomaash:
This tests whether all nodes are placed and clickable (after stabilize and fit), the correct number of edges connects the visible nodes and that levels are evenly spaced without gaps. All forementioned is tested again before and after clustering. The test checks linear graphs, trees and general acyclic graphs. Some tests are also performed agains simple cyclic graphs but that's mostly to just make sure the layouting algorithm doesn't cycle forever or something like that.

This commit comes with universal HTML page that exposes it's variables for the spec file to use and commands that can take advantage of this. The idea here is to be able to execute any code from the spec file without having to create a new HTML for it. This was created as a side effect but in the end it is probably way more useful than the test itself.

I tried the test against older versions of the library and it failed due to nodes being placed in wrong levels or gaps being created after clustering. In other words it detects regressions it is supposed to detect.

### DIFF
--- a/cypress/integration/directed-hierarchical-layout.spec.ts
+++ b/cypress/integration/directed-hierarchical-layout.spec.ts
@@ -1,0 +1,324 @@
+import { IdType, TestData, addMoreEdges, generateMaryTree } from "./helpers";
+
+/**
+ * Generate a consecutive list of ids starting with first and ending with last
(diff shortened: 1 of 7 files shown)


Thomaash:
Thanks.

PR #334, opened 2020-01-08
feat(physics): add wind

ylarom:
In each physics step (_performStep) after calculating all forces on the node, an additional force is applied.
This causes all nodes to be pushed in the given direction.
You can see a sample of the effects in physicsConfiguration.html using Forced Atlas 2 Base solver (in the other solvers the entire network drifts away since there are no fixed nodes so nothing pulls the network to (0,0))

### DIFF
--- a/docs/network/physics.html
+++ b/docs/network/physics.html
@@ -132,7 +132,8 @@ var options = {
       fit: true
     },
     timestep: 0.5,
-    adaptiveTimestep: true
+    adaptiveTimestep: true,
+    wind: { x: 0, y: 0 }
   }
 }
 
@@ -197,6 +198,11 @@ network.setOptions(options);
         <tr parent="stabilization" class="hidden"><td class="indent">stabilization.fit</td>                      <td>Boolean</td>        <td><code>true</code></td>       <td>Toggle whether or not you want the view to zoom to fit all nodes when the stabilization is finished.</td></tr>
         <tr><td>timestep</td>                               <td>Number</td>         <td><code>0.5</code></td>        <td>The physics simulation is discrete. This means we take a step in time, calculate the forces, move the nodes and take another step. If you increase this number the steps will be too large and the network can get unstable. If you see a lot of jittery movement in the network, you may want to reduce this value a little.</td></tr>
         <tr><td>adaptiveTimestep</td>                       <td>Boolean</td>        <td><code>true</code></td>       <td>If this is enabled, the timestep will intelligently be adapted <b>(only during the stabilization stage if stabilization is enabled!)</b> to greatly decrease stabilization times. The timestep configured above is taken as the minimum timestep. <a href="layout.html#layout" target="_blank">This can be further improved by using the improvedLayout algorithm</a>.</td></tr>
+
+        <tr class='toggle collapsible' onclick="toggleTable('optionTable','wind', this);"><td><span parent="wind" class="right-caret"></span> wind</td>                                                     <td>Object</td><td><code>Object</code></td>    <td>A force that pushes all non-fixed nodes in the given direction.<br/>Requires all nodes are connected to nodes which are `fixed`, otherwise non-attached nodes will keep moving indefinitely.</td></tr>
+        <tr parent="wind" class="hidden"><td class="indent">wind.x</td>     <td>Number</td>         <td><code>0</code></td>       <td>The amount of force to be applied pushing non-fixed nodes to the right (positive value) or to the left (negative value).</td></tr>
+        <tr parent="wind" class="hidden"><td class="indent">wind.y</td>     <td>Number</td>         <td><code>0</code></td>       <td>The amount of force to be applied pushing non-fixed nodes downwards (positive value) or upwards (negative value).</td></tr>
+
     </table>
 
     <div class="flagbar">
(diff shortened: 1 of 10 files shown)


docs/network/physics.html
```diff
@@ -197,6 +198,11 @@ <h3>Options</h3>
...
+
+        <tr class='toggle collapsible' onclick="toggleTable('optionTable','wind', this);"><td><span parent="wind" class="right-caret"></span> wind</td>                                                     <td>Object</td><td><code>Object</code></td>    <td>A force that pushes all non-fixed nodes in the given direction.<br/>Requires all nodes are connected to nodes which are `fixed`, otherwise non-attached nodes will keep moving indefinitely.</td></tr>
+        <tr parent="wind" class="hidden"><td class="indent">wind.x</td>     <td>Number</td>         <td><code>0</code></td>       <td>The amount of force to be applied pushing non-fixed nodes to the right (positive value) or to the left (negative value).</td></tr>
+        <tr parent="wind" class="hidden"><td class="indent">wind.y</td>     <td>Number</td>         <td><code>0</code></td>       <td>The amount of force to be applied pushing non-fixed nodes downwards (positive value) or upwards (negative value).</td></tr>
+
```
Thomaash: Could you also add these to the Korean docs, please? If you don't speak Korean just replace the English texts with `TODO` so that we know there's something missing.

lib/network/modules/LayoutEngine.js
```diff
@@ -505,6 +505,9 @@ class LayoutEngine {
           }
         }
       }
+      if (allOptions.physics && !allOptions.physics.wind) {
+        allOptions.physics.wind = { x: 0, y: 0 };
+      }
```
Thomaash: What is the purpose of this? Reading through the code I just don't see anything this would change? Plus why is layout engine involved in physics? Does it interfere with something? If so there definitely should be a comment explaining why this is.

lib/network/options.js
```diff
@@ -670,6 +675,10 @@ let configureOptions = {
     minVelocity: [0.1, 0.01, 0.5, 0.01],
     solver: ['barnesHut', 'forceAtlas2Based', 'repulsion', 'hierarchicalRepulsion'],
     timestep: [0.5, 0.01, 1, 0.01],
+    wind: {
+      x: [0, -10, 10, 0.1],
+      y: [0, -10, 10, 0.1]
+    },
```
Thomaash: Playing with this it seems that 2 is already a really strong wind, 10 is insane (the nodes just fly off the screen immediately and the edges get really stretched out). I think this would be much better with smaller range and finer steps. I think most people would use wind +-0.x anyway.

PS: As always I may be wrong. If these strong winds have valid use cases feel free to correct me.
PPS: It would be nice to have an example with some fixed nodes and wind sliders to play with. In the existing physics example it just drifts off the screen.

ylarom:
Re the e2e tests - they sometimes pass and sometime fail on my machine
Could there be a random element there?
(btw, is there a command line option to i run a specific test?)

docs/network/physics.html
ylarom: Added the rows with a TODO placeholders in the Korean doc

Thomaash:
If the tests fail randomly you should open an issue for that.

Regarding running specific tests you can edit the test file and add `.only` (for example `it('test name'...` to `it.only('test name'...`) to run only that test. There is also a way how ti pick them based on regexp but I don't remember how to invoke it (try `--help` 😄). Cypress also has GUI that lets you choose specific test files: `npm run test:e2e:gui`.

lib/network/options.js
ylarom: The size of the "wind" force required to move nodes is related to countering the forces applied by the "spring" edges pulling from fixed nodes.
In a network with three nodes, a 0.1 wind will do.
When there are more fixed nodes, a stronger force is needed.

Added a "wind.html" under examples/physics

lib/network/modules/LayoutEngine.js
ylarom: Removed

ylarom:
Sure
There are no validations right now in setOptions, so I don't know how you
guys would like it
Will this do:
[image: image.png]

Thanks

Thomaash:
Hi @ylarom,

sorry for the delay. I thought I answered this right away but I probably forgot to send it and closed the tab or something. Anyway the screenshot didn't get through, there's just `[image: image.png]`. Could you just copy paste the code here or even better just push it into the repo so I can review it?

Thanks.

lib/network/modules/PhysicsEngine.js
```diff
@@ -162,6 +162,18 @@ class PhysicsEngine {
...
+          const wind = this.options.wind;
+          if (wind) {
+            if (wind.x && typeof wind.x !== 'number' || isNaN(wind.x)) {
+              delete wind.x;
+            }
+            if (wind.y && typeof wind.y !== 'number' || isNaN(wind.y)) {
+              delete wind.y;
+            }
+          }
+        }
```
Thomaash: I'm not saying this wouldn't work because it would but:
- `wind.x && ` doesn't do anything there.

ylarom:
Thank you!

PR #335, opened 2020-01-08
Fix labels in Chosen Elements example

kmalyavina:
The labels in the [Chosen Elements example](https://visjs.github.io/vis-network/examples/network/other/chosen.html) do not change on hover as they're supposed to. Using the `ctx` param instead of `values` when editing the properties fixes the problem.

### DIFF
--- a/examples/network/other/chosen.html
+++ b/examples/network/other/chosen.html
@@ -57,32 +57,32 @@
 
   var changeChosenLabelColor =
     function(ctx, values, id) {
-      values.color = "#ff0000";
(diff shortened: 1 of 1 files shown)


Thomaash:
Hi @kmalyavina,

thanks for your effort but I don't think this is a bug in the example. I think this is a bug in the library. I'll have to investigate this some more to be sure though.

Thomaash:
Oh it is an error in the example after all. Because the first parameter is values not ctx. The function doesn't receive ctx at all.

PR #350, opened 2020-01-12
chore(examples): update conf for new Vis Dev Utils

Thomaash:
This updates (.gitignore, package.json) and creates (generate-examples-index.json) everything necessary for the new version.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,7 +31,9 @@ coverage/
 /cypress/videos/
 /declarations/
 /dist/
+/examples/examples.css
 /examples/index.html
+/examples/static/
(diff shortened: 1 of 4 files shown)


yotamberk:
Can you do this for vis-timeline as-well?

Thomaash:
Timeline already has the config files (see <https://github.com/visjs/vis-timeline/blob/master/generate-examples-index-timeline.json> and <https://github.com/visjs/vis-timeline/blob/master/generate-examples-index-graph2d.json>). You can simply update the package as the config file format is unchanged since the last time.

PR #352, opened 2020-01-13
fix(rng)!: use better seedable RNG

Thomaash:
BREAKING CHANGE: The layout for the same graph and seed is different
when compared to the previous version.

This replaces the Math.sin() based ones and Math.random() with Alea
based RNG from Vis Util.

A few pinning tests (the main reason why Alea was introduced) are
included.


test/snapshots.test.ts
```diff
@@ -0,0 +1,200 @@
...
+        { from: 8, to: 16 },
+        { from: 8, to: 17 },
+        { from: 9, to: 18 },
+        { from: 9, to: 19 }
+      ]
+    }
+  }
+];
+
+describe.only("Physics snapshots", function(): void {
```
yotamberk: remove `.only`

Thomaash: Good thing that I forgot this here because enabling the other tests changed the results. This should definitely not happen. I'll have to look deeper into this because so far it is a mystery to me how this can even happen.

Thomaash: I found a way around this but I'm afraid pinning tests are going to be very fragile. I'll try it out and see. It can always be removed later so no big deal.

Thomaash:
I run all changes through Prettier which keeps it from exceeding 80 columns. In case of long URLs in script tags they end up on three lines.

Issue #201, opened 2019-11-01
Half circle of self reference option


piratuks:
I would like to request features / properties related to the selfReference. Property to say that selfReference edge should be half circle. I have network where i am representing some nodes as tables and i have other nodes as attributes of the table (meaning these are placed inside that table node). So in this case i have table node with transparent background. Now if such table has self reference half of the edge is fine because it is above node and other half is visible under the node which is a bit annoying.

I am attaching images to represent my requested option:

So the following image is something i would expect from requested options. You can see that the table is transparent because other edge is visible, however self ref is just a half of it.

![Capture](https://user-images.githubusercontent.com/8760197/68024179-02a11480-fcb2-11e9-9f24-7e0c30a70bbd.PNG)

Thomaash:
Implemented with #368.

PR #368, opened 2020-01-21
Half circle of self reference option

piratuks:
Implementing additional properties startAngle and endAngle which helps to define self reference circle size (half circle and etc). Example is under selfReference.html #201


piratuks:
@Thomaash  i see your point. Regarding the switch approach. So let me clarify you suggest flag property instead of a startAngle, endAngle properties which indicates that the part behind parent node should not be rendered. Which basically means to calculate startAngle and endAngle our selfs in the code. Is that correct? If it is could you give me a helping hand because right now i am thinking on proper way to calculate these angles.

Thomaash:
You understand it correctly. Regarding help I don't have much time right now but have a look at the arrows. You basically want to draw an arc from the from arrow through the middle arrow to the to arrow so the positions of these arrows should also tell you where to start and stop the arc.

piratuks:
@Thomaash Thanks for quick guiding help. I took the approach. Changed properties to the flag and calculating start/end angle. Please check the approach would you agree on this.

examples/network/edgeStyles/selfReference.html
```diff
@@ -27,23 +27,99 @@
...
-                y:200
+                label: "Test shape",
+                widthConstraint: { minimum: 100 },
+                heightConstraint: { minimum: 100 },
+                x: 400,
+                y: 400
+            },
+            {
+                id: 3,
+                label: "This is another test shape.",
```
Thomaash: It would be great to use the labels to describe what configuration was used to achieve this effect.

examples/network/edgeStyles/selfReference.html
```diff
@@ -72,10 +148,16 @@
                         }
                     }
                 });
+
+                edges.updateOnly({
+                    id: 3,
+                    selfReference: {
+                        fullRender: false
+                    }
+                });
```
Thomaash: You set the same thing over and over again. Configure it once when the edge is initially created.

types/network/Network.d.ts
```diff
@@ -971,6 +971,7 @@ export interface EdgeOptions {
   selfReference?: {
     size?: number,
     angle?: number,
+    fullRender?: boolean
```
Thomaash: I don't like the `fullRender` name. It's really vague and doesn't give a good idea about the feature it controls. The only thing that comes to my mind right now is `renderBehindTheNode`. Although it is more functional it's not very elegant. Feel free to suggest something else.

piratuks:
@Thomaash Changed example .html. Regarding naming huh thats hard, changed to your suggested, but maybe something like overlapingPartsOn, renderOverlapingParts (however this one similar to your suggestion),

examples/network/edgeStyles/selfReference.html
```diff
@@ -26,100 +26,30 @@
             {
                 id: 1,
                 label:
-                    "I'm an alligator,\nI'm a mama-papa coming for you.\nI'm the space invader,\nI'll be a rock 'n' rollin' bitch for you",
+                    "Node has selfReference edge,\nwhich angle (position) is provided.\nProperty renderBehindTheNode allows to draw parts,\nwhich are not overlaping with the node",
```
Thomaash: A code readability tip here: You can use the following pattern instead of having a really long string with arbitrarily placed `\n`s:
```javascript
const multilineString = [
  "The first line goes here.",
  "The second line comes right bellow it.",
  "Here goes the third.",
  "Finally the fourth and last line."
].join("\n");
```

lib/network/modules/components/Edge.js
```diff
@@ -256,7 +256,7 @@ class Edge {
       parentOptions.selfReference = {
           size: newOptions.selfReferenceSize,
           angle: parentOptions.selfReference.angle,
-          fullRender: parentOptions.selfReference.fullRender
+          renderBehindTheNode: parentOptions.selfReference.renderBehindTheNode
       }
```
Thomaash: Can't you just change the `size` here instead of creating a new object? I mean if `selfReference` isn't an object it will crash anyway so what's the difference? Can't it crash btw?

docs/network/edges.html
```diff
@@ -927,7 +927,7 @@ <h3>Options</h3>
             </td>
         </tr>
         <tr parent="selfReference" class="hidden">
-            <td class="indent">selfReference.fullRender</td>
+            <td class="indent">selfReference.renderBehindTheNode</td>
             <td>Boolean</td>
             <td><code>true</code></td>
             <td> This is the flag to indicate if self reference circle should be drawn as a full circle or rather should not overlap with parent node.
```
Thomaash: `This is a flag that indicates whether` sound better.

examples/network/edgeStyles/selfReference.html
```diff
@@ -22,34 +22,73 @@
...
+            "I'm the space invader,",
+            "I'll be a rock 'n' rollin' bitch for you."
+        ].join("\n");
+
         var nodes = new vis.DataSet([
             {
                 id: 1,
                 label:
-                    "Node has selfReference edge,\nwhich angle (position) is provided.\nProperty renderBehindTheNode allows to draw parts,\nwhich are not overlaping with the node",
+                    multilineString1,
```
Thomaash: I think this should've been multilineString2. Also the variable name is just lazy and since it's used just once inline it here, there's no need for a variable in this case.

examples/network/edgeStyles/selfReference.html
```diff
@@ -22,34 +22,73 @@
...
-                 color: {
+                color: {
                     background: "transparent"
                 }
             },
             {
                 id: 2,
                 label:
-                    "I'm an alligator,\nI'm a mama-papa coming for you.\nI'm the space invader,\nI'll be a rock 'n' rollin' bitch for you",
+                    multilineString2,
```
Thomaash: I imagined something like:

> An edge connected to the same node on both end renders a circle.
> Both from and to arrows point to the same node.
> The radius of the edge can be adjusted using selfReference.size.
> The edge can be positioned using selfReference.angle.
> The part of the edge behind the node can be hidden using selfReference.renderBehindTheNode.

examples/network/edgeStyles/selfReference.html
```diff
@@ -22,34 +22,73 @@
...
+                shape: "box"
+            },
         ]);
 
         // create an array with edges
         var edges = new vis.DataSet([
             { from: 1, to: 1, label: "Moonage Daydream" },
-            { from: 2, to: 2, label: "Testing", arrows: 'to, middle, from', selfReference: { size: 40, angle: Math.PI / 6 } }
+            {
+                from: 2, to: 2, label: "Testing", arrows: 'to, middle, from',
```
Thomaash: > selfReference.renderBehindTheNode: false

examples/network/edgeStyles/selfReference.html
```diff
@@ -22,34 +22,73 @@
...
+                selfReference: {
+                    size: 40,
+                    angle: Math.PI / 6
+                }
+            },
+            {
+                from: 3, to: 3, label: "Testing", arrows: 'to, middle, from',
```
Thomaash: > selfReference.renderBehindTheNode: true

piratuks:
@Thomaash updated example .html

Thomaash:
Oh, @piratuks,

one more thing. Could you update your branch to make sure everything still works with the recent changes from master?

Thanks.

piratuks:
@Thomaash  ye give me couple minutes

piratuks:
@Thomaash Merged master to the branch. No conflicts. Performed a build and checked the .html looks okay.

Thomaash:
Thanks. It'll be released soon.

Just a note regarding your English: You can never omit the subject in English sentences. For example your last comment should have been `@Thomaash, I merged...`. The way you wrote it reads as if you were saying that I merged it.

Issue #365, opened 2020-01-20
Label orientation/position around node


kmalyavina:
I would like to be able to put the label to the right of a node. My use case is adding a legend to the graph. Label-centered-at-the-bottom gets quickly takes up a lot of vertical space and gets hard to follow when there are a lot of groups to show. 

I've tried a number of ways with margin, font alignment, circle-physics-off-with-small-size but could not get the text to line up properly. 

Idealy, you would be able to choose top/bottom/left/right for the position of the node label (in an attribute like "labelPosition")

Current -> Desired (example from OrientDB Studio)
<img width="349" alt="Screen Shot 2020-01-20 at 8 25 19 AM" src="https://user-images.githubusercontent.com/22348821/72729987-6f63d680-3b5e-11ea-8b30-876c56143dbf.png">

Thomaash:
Hi @kmalyavina,

would you have the time to work on this?

kmalyavina:
> Hi @kmalyavina,
> 
> would you have the time to work on this?

I'm not very familiar with the codebase, but I might be able to try implementing it if given some guidance.

Thomaash:
That would be great. Feel free to open a draft PR once you have anything even if it doesn't work (it's much easier to offer guidance while looking at the code).

The place to start are the files `lib/network/modules/components/Node.js` which invokes the label rendering and maintains info about it's shape and position and `lib/network/modules/components/shared/Label.js` which renders the label itself.

PR #373, opened 2020-01-22
Draft for Node Label Position

kmalyavina:
For #365 

Works fine for default size nodes with constant values for x/y offsets. Need to include the node radius/size in the offset instead. 

I have a couple questions and thoughts/suggestions. 

Questions:
1) In the label.size, what is the yLine?
2) How does it currently get the y value/distance to position the label below a node?
3) How can I get the node width/height to use as offset for label alignment?

Thoughts/Suggestions:
* Based on the usage and contents of the object, 'dimensions' seems a more appropriate name than 'size' (worried about changing it in case it's used elsewhere)
* The whole network body is passed in for label creation, but it looks like only view.scale is used - why not just pass in the scale?
* Positioning label given center/origin of a node would probably make more sense than adjusting based on current position at bottom (related to question 2 above)

### DIFF
--- a/lib/network/modules/components/Node.js
+++ b/lib/network/modules/components/Node.js
@@ -511,7 +511,7 @@ class Node {
 
 
   /**
-   * check if this node is selecte
+   * check if this node is selected
    * @return {boolean} selected   True if node is selected, else false
    */
   isSelected() {
--- a/examples/network/labels/labelAlignment.html
+++ b/examples/network/labels/labelAlignment.html
@@ -20,10 +20,9 @@
 
 <body>
 
-<p>Labels of edges can be aligned to edges in various ways.</p>
-<p>Text-alignment within node labels can be 'left' or 'center', other font alignments not implemented.</p>
-<p>Label alignment (placement of label &quot;box&quot;) for nodes (top, bottom, left, right, inside) is
-planned but not in vis yet.</p>
+<p>Labels of edges can be aligned to edges either: <b>horizontal</b>, middle, top, or bottom.</p>
+<p>Text-alignment within node labels can be aligned: <b>center</b>, left, or right.</p>
+<p>For dot-like nodes, labels can be positioned at: <b>bottom</b>, inside, top, left, or right.</p>
 <p>The click event is captured and displayed to illustrate how the clicking on labels works.
 You can drag the nodes over each other to see how this influences the click event values.
 </p>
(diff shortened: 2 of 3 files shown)


Thomaash:
Hi there,

thanks for working on this. I don't have much time right now and most likely won't have tomorrow. I'll have a look probably at the weekend.

PS: If you won't hear from me by the next week feel free to mention me here to remind me.

Thomaash:
First of all thanks for the example. It's really nice to just check it out, run it and see it in action. The code you have so far also looks good given it's surrounding.

### Answers
1. I have no idea what yLine is. I looked through the code but I'm as clueless as you.
1. The logic seems to be split between several methods and seems to be partly computed outside of the Label class. The interesting place here are the various draw methods of the shapes in `lib/network/modules/components/nodes`. In case you wonder this design pattern is commonly referred to as spaghetti code.
1. The node has a shape property which has bounding box property (`node.shape.boundingBox`) or it is possible to get much more accurate data for any specific angle but if the choices are top, right, bottom, left and inside there's no point in doing that.

### Renaming size to dimensions
This seems much more descriptive. It's being used at https://github.com/visjs/vis-network/blob/7d3361c91713898f4dd04b74e37d737358c6cf2e/lib/network/modules/components/Edge.js#L783 and there is also very interesting way of using it at https://github.com/visjs/vis-network/blob/7d3361c91713898f4dd04b74e37d737358c6cf2e/lib/network/modules/components/Node.js#L537 (Yes, this will throw when invoked. I think it's safe to just delete this one.)

### Network body
The scale can change at any point so it has to be provided within an object. The body is a single object maintaining the state of the whole Network used all over the place so it's probably best left as it is if just for the sake of consistency.

### Positioning the label
The way it is doesn't seem to make any sense whatsoever. What would maybe make more sense would be if the label class handled just the label itself, provided it's bounding box to the node and the positioning would be done by the node.

### Afterthought
Going through the code maybe it would be best to throw it away and rewrite the whole thing from the scratch. This is just unmaintainable mess. Though who has the time to do something like that?

kmalyavina:
Haha, rewriting the whole thing from scratch would be pretty insane. 

I think I at least figured out what "yLine" is - it's basically the vertical middle of the label. 

I'll attempt to move all the label positioning parts to ShapeBase and see if I can get it a little better at least.

kmalyavina:
Feels like I didn't do much to help the spaghetti code problem, but I at least got the node placement to work as intended.

Big YIKES at all the duplicate`updateBoundingBox`. Trying to update them to make them accurate to take into account labels on the top/sides.

lib/network/modules/components/Node.js
```diff
@@ -527,17 +527,15 @@ class Node {
...
   /**
    * Get the current dimensions of the label
    *
    * @return {rect}
    */
-  getLabelSize() {
-    return this.labelModule.size();
+  getLabelDimensions() {
+    return this.labelModule.getDimensions();
   }
```
Thomaash: Unless you need it delete it. The only thing it was doing before was throwing errors. Nobody complained so nobody needs it.

lib/network/modules/components/nodes/util/ShapeBase.js
```diff
@@ -68,9 +68,33 @@ class ShapeBase extends NodeBase {
...
       // Need to call following here in order to ensure value for `this.labelModule.size.height`
-      this.labelModule.calculateLabelSize(ctx, selected, hover, x, y, 'hanging')
-      let yLabel = y + 0.5 * this.height + 0.5 * this.labelModule.size.height;
-      this.labelModule.draw(ctx, x, yLabel, selected, hover, 'hanging');
+      this.labelModule.calculateLabelDimensions(ctx, selected, hover, x, y, 'hanging'); 
+
+      let labelX = x;
+      let labelY = y;
+      let labelBaseline = 'middle';
+      const DISTANCE = 12; // note: 12 is maybe the line height??
```
Thomaash: Font size can be changed and with it line height.

lib/network/modules/components/nodes/util/ShapeBase.js
```diff
@@ -68,9 +68,33 @@ class ShapeBase extends NodeBase {
...
+
+      switch(this.labelModule.position){ 
+        case 'inside': 
+          break; 
+        case 'left': 
+          this.labelModule.fontOptions.align = 'right'; // set font alignment to right
```
Thomaash: This should be done when the options are being processed in `setOptions`. Though a question comes to my mind whether it is necessary. If someone wants the text to be aligned away from the node then just let them do it.

lib/network/modules/components/nodes/util/ShapeBase.js
```diff
@@ -68,9 +68,33 @@ class ShapeBase extends NodeBase {
...
+
+      switch(this.labelModule.position){ 
+        case 'inside': 
+          break; 
+        case 'left': 
+          this.labelModule.fontOptions.align = 'right'; // set font alignment to right 
+          labelX -= (this.labelModule.dimensions.width + this.width) * 0.5 + DISTANCE; // shift left 
+          break; 
+        case 'right': 
+          this.labelModule.fontOptions.align = 'left'; // set font alignment to left
```
Thomaash: dtto

kmalyavina:
Hmm... when I go back to before renaming size to dimensions ("Label position initial draft" commit), I don't have the selecting/dragging issues. I'll try to go back to that point and redo the changes from that point...

kmalyavina:
Ok, got those changed merged in after undoing the 'dimensions' rename. 

I noticed a weird thing though that looks like all networks are shifted up?? Not sure what could be causing it

lib/network/modules/components/nodes/util/ShapeBase.js
```diff
@@ -82,17 +106,19 @@ class ShapeBase extends NodeBase {
...
    */
   updateBoundingBox(x, y) {
+    // TODO: Same/similar code is at least 4 places!! 
     this.boundingBox.top = y - this.options.size;
     this.boundingBox.left = x - this.options.size;
     this.boundingBox.right = x + this.options.size;
     this.boundingBox.bottom = y + this.options.size;
 
     if (this.options.label !== undefined && this.labelModule.size.width > 0) {
+      this.boundingBox.top = Math.max(this.boundingBox.top, this.labelModule.size.top);
```
Thomaash: You have `Math.max` here and `Math.max` at the bottom too. I think this one should be `Math.min` to include the label in the bounding box. I don't think this will solve all the issues though.

kmalyavina:
Spent a lot of time trying to fix that initial load without much luck, and I'm at the end of the time I can dedicate to working on this :(

Issue #202, opened 2019-11-01
Ability to set edge angle / length


piratuks:
I would like to request features / properties related to the selfReference. Property / option / ability to say that edge between two nodes should be at a specific angle or length. Because right now if i have table which attributes are pointing to other attribute then the edge will always be inside table node...

I am attaching images to represent my requested option. i would expect ability to set edge between nodes (in this case attributes) on a specific length / angle something like below. 

![Capture1](https://user-images.githubusercontent.com/8760197/68024340-7fcc8980-fcb2-11e9-9dd7-e0e1fe9a064d.PNG)

PR #376, opened 2020-01-23
Ability set edge length

piratuks:
For #202 

Just started on this. However, i have in mind to provide some spacing which would affect cubicBezier. Actually already did for it. However i imagine that's not quite correct since there is option under smooth which affects only one type of edge. 

Could you spill out your idea on how that suppose to look for end result?


Thomaash:
How exactly does this differ from the edge length option we already have?

piratuks:
> 
> 
> How exactly does this differ from the edge length option we already have?

@Thomaash  Maybe i expressed my self incorrectly. So the length you already have is basically distance between x1,y1 to x2,y2. I am talking about additional property which would define angle / offset. I am not sure if that is already possible but from what i have seen it is is not.

I attache quick drawing: so you have two circle shapes. Edge between them could be like in the drawing and the angle how big it is based on provided property.

Like in the issue itself (image for the edge arch)


test

Thomaash:
Okay that makes a lot more sense but wouldn't it be better to create a new arch edge type for this rather than augmenting existing one?

piratuks:
@Thomaash  So i am providing radius and type which i called in this demo (arc). Now i am a bit struggling to calculate radius. So currently i hardcoded to 2 * Math.PI which results in the 
Capture1

Now when position of nodes changed (dragged) then it would be cool to adjust the angle because right now it is hardcoded so it obviously is not working but my aim is to achieve the following: 
Capture2

And so on... Meaning changing position keeps the arc shape.

So i am a bit struggling to calculate that angle instead of providing it via property or hardcoding. Could you suggest something?

lib/network/options.js
```diff
@@ -598,9 +599,10 @@ let configureOptions = {
     },
     smooth: {
       enabled: true,
-      type: ['dynamic', 'continuous', 'discrete', 'diagonalCross', 'straightCross', 'horizontal', 'vertical', 'curvedCW', 'curvedCCW', 'cubicBezier'],
+      type: ['dynamic', 'continuous', 'discrete', 'diagonalCross', 'straightCross', 'horizontal', 'vertical', 'curvedCW', 'curvedCCW', 'cubicBezier', 'arc'],
       forceDirection: ['horizontal', 'vertical', 'none'],
-      roundness: [0.5, 0, 1, 0.05]
+      roundness: [0.5, 0, 1, 0.05],
+      radius: [0,100,200,400],
```
Thomaash: The radius is `[default, min, max, step]`. Having the default outside the range and the step four times the range renders this unusable.

piratuks: Thats the thing as similar. The difference is just that i want to provide angle / length to get results as above in the image. Thats the part i am a bit struggling with.

Thomaash: Okay.

What you want to play with is `Math.atan2(dy, dx)`. This will give you the angle from one node to the other. But you also have to compute `dy` and `dx` the same way. Computing one as from - to and the other as to - from is just not going to work.

PR #399, opened 2020-01-28
Hover self ref

### LINKED ISSUE #398: hover self ref
Also found that when you try to hover the edge which angle is not Math.PI / 4 then it fails. I found that it is because on however method getDistanceToEdge is called and it executes this._getCircleData(undefined);. Basically passed ctx is undefined which means getSelfRefCoordinates - >if(ctx && typeof node.distanceToBorder === "function"){ returns incorrect data.  

So the solution could be to remove case sentence and check under each _distanceToBorder if ctx was provided, basically skip resize(ctx), because in this case it is needed just calculation.

Any thoughs?

piratuks:
Closes #398 
Hotfix -> hover is getting incorrect position due to not providing ctx


Thomaash:
Could you just update the branch so that there are no nasty surprises upon release? Thanks.

piratuks:
@Thomaash yea updated.

PR #402, opened 2020-01-28
Render behind node

### LINKED ISSUE #397: render behind node
I found issue when i added to my project. Only for specific case -> shape must be box or text and heightConstraint between 175 - 431. Then the angleFrom and To are calculated wrongly.

Have i missed something @Thomaash ?

        var nodes = new vis.DataSet([
            {
                id: 2,
                label:
                    [
                        "An edge connected to the same node on both end renders a circle.",
                        "Both from and to arrows point to the same node.",
                        "The radius of the edge can be adjusted using selfReference.size.",
                        "The edge can be positioned using selfReference.angle.",
                        "The part of the edge behind the node can be hidden using selfReference.renderBehindTheNode."].join("\n"),
                 heightConstraint: { minimum: 400 },
                x: 100,
                y: 0,
                shape: "box",
                color: {
                    background: "transparent"
                }
            }, 
        ]);

        // create an array with edges
        var edges = new vis.DataSet([
            { from: 1, to: 1, label: "Moonage Daydream" },
            {
                from: 2, to: 2, label: "Testing", arrows: 'to, middle, from',
                selfReference: {
                    size: 30,
                    angle: Math.PI / 2,
                    renderBehindTheNode: false
                }
            },

        ]);
    
        // create a network
        var data = {
            nodes: nodes,
            edges: edges
        };
        var options = {};

piratuks:
fixes #397 #201

Changed searching for an intersection


piratuks:
@Thomaash Do you agree on change?

lib/network/modules/components/edges/util/edge-base.ts
```diff
@@ -535,15 +535,17 @@ export abstract class EdgeBase<Via = undefined> implements EdgeType {
       //render only parts which are not overlaping with parent node
       //need to find x,y of from point and x,y to point
       //calculating radiangs
-      const low = this.options.selfReference.angle - 2 * Math.PI;
-      const high = this.options.selfReference.angle;
+      let low = this.options.selfReference.angle - 0.1;
+      let high = this.options.selfReference.angle + Math.PI;
```
Thomaash: What about the magical constant 0.1? Why not start from the angle and check a whole half of the circle? If there's a reason there definitely should be a comment explaining why the angle can't be used and how the 0.1 makes it work.

Also I think you can simply inline these values. Each is used only once so there's no reason to create and mutate a variable. It's just weird.

piratuks: The idea for 0.1 was to make start angle slightly smaller that when it is searched for intersection it would always find different one. However @Thomaash you are right it looks weird. Changed to angle / angle + pi

PR #554, opened 2020-03-19
Node Opacity

Tyler-Maclachlan:
Hey I added a node option for overall node opacity (it will set border, shadows and images).

I also added ordering for hovered nodes in the renderer for: https://github.com/visjs/vis-network/issues/226


Tyler-Maclachlan:
Hey @Thomaash @mojoaxel 

Can you guys give this a look

lib/network/modules/components/Node.js
```diff
@@ -245,15 +247,16 @@ class Node {
...
     var fields = [
       'color',
+      'opacity',
       'fixed',
       'shadow'
     ];
     util.selectiveNotDeepExtend(fields, parentOptions, newOptions, allowDeletion);
 
     Node.checkMass(newOptions);
+    parentOptions.opacity = newOptions.opacity || parentOptions.opacity || 1;
```
Thomaash: Do NOT modify group options from here! This way you'll set this nodes opacity to the group. Instead the group's opacity should be copied into the node if it doesn't have it's own.

lib/network/modules/components/nodes/shapes/Image.js
```diff
@@ -1,7 +1,7 @@
 'use strict';
 
 import CircleImageBase from '../util/CircleImageBase'
-
+import * as util from 'vis-util/esnext';
```
Thomaash: Don't import everything. Import only what you need.

Tyler-Maclachlan:
Hey @Thomaash 

I've added an example and fixed up the things.

Tyler-Maclachlan:
Hey @Thomaash 

Fixed the issues, they were all because of a falseness check.

Tyler-Maclachlan:
Oops, clicked close and comment by accident

Tyler-Maclachlan:
- NaN, infinity etc. should be ignored and reported (console.error). Now it misbehaves with them. - done
 
- Values outside of <0, 1> should also be ignored and reported. - done
 
- Explicitly set opacity on a node seems to be ignored when the node is selected. - I have one in the example and it's working as it should, can you provide a codepen?
 
- The example should say something more than “Display nodes as images”. - done
 
- Your branch is out of date. It's necessary to merge changes from the master before merging into the master. - done

lib/network/modules/NodesHandler.js
```diff
@@ -180,9 +180,10 @@ class NodesHandler {
     if (options !== undefined) {
       Node.parseOptions(this.options, options);
       
+      // Need to set opacity here because parseOptions-parentOptions can't be overridden incase of node being in a group
```
Thomaash: I don't understand this comment.

lib/network/modules/components/Node.js
```diff
@@ -258,8 +267,20 @@ class Node {
     util.selectiveNotDeepExtend(fields, parentOptions, newOptions, allowDeletion);
 
     Node.checkMass(newOptions);
-    if (globalOptions.opacity !== undefined && newOptions.opacity === undefined) {
-      newOptions.opacity = globalOptions.opacity;
+
+    
+    if (parentOptions.opacity !== undefined) {
+      if (Number.isNaN(parentOptions.opacity) || !Number.isFinite(parentOptions.opacity) || parentOptions.opacity < 0 || parentOptions.opacity > 1) {
```
Thomaash: This insanity can actually be simplified to:
```javascript
!(0 <= opacity && opacity <= 1)
```

lib/network/modules/components/Node.js
```diff
@@ -258,8 +267,20 @@ class Node {
...
+        console.error("Invalid option for node opacity. Value must be between 0 and 1, found: " + parentOptions.opacity);
+        parentOptions.opacity = undefined;
+      }
+    }
+
+    if (newOptions.opacity !== undefined) {
+      if (Number.isNaN(newOptions.opacity) || !Number.isFinite(newOptions.opacity) || newOptions.opacity < 0 || newOptions.opacity > 1) {
+        console.error("Invalid option for node opacity. Value must be between 0 and 1, found: " + newOptions.opacity);
+        newOptions.opacity = undefined;
+      }
```
Thomaash: It would be great to have this in a function rather than copy pasting it all over the place.

lib/network/modules/components/Node.js
Tyler-Maclachlan: Created a function called `checkOpacity`

lib/network/modules/NodesHandler.js
Tyler-Maclachlan: Cleared this comment up

Tyler-Maclachlan:
This wasn't a leaking of options between nodes. It was the fact that the context was being changed without being restored.

lib/network/modules/components/Node.js
```diff
@@ -139,10 +140,18 @@ class Node {
...
     Node.parseOptions(this.options, options, true, this.globalOptions, this.grouplist);
-
-    if (options.opacity !== undefined) {
+    
+    
+    if (options.opacity !== undefined && Node.checkOpacity(options.opacity)) {
+      if (this.id >= 4 && this.id <= 7) {
+        console.log(options.opacity);
+      }
+      // FIXME: This doesn't work if a group has already set this option
```
Tyler-Maclachlan: I need some help with the options, I can't seem to find why the group opacity overrides the setting of the opacity here

Tyler-Maclachlan: Fixed it

Tyler-Maclachlan:
Too many functions take options and overwrite them. Options should be read only. 

Check out how [chartjs](https://github.com/chartjs/Chart.js/blob/master/src/core/core.defaults.js) does theirs.

Thomaash:
@Tyler-Maclachlan I'd like to rework the options (in fact I already did a lot of work on this), but the problem is that the options are so complex and I don't have the time to finish it. You can see my work at https://github.com/visjs/vis-util/pull/71, just in case your interested.

In the unlikely case you would have the time to do a complete overhaul of this, that would be incredibly awesome.

PR #576, opened 2020-03-26
fix(hierarchical-repulsion-solver): remove assignment to const

### LINKED ISSUE #555: Bug in solver() in HierarchicalRepulsionSolver.js:70
OSX 10.14.5
Latest visjs from https://unpkg.com/vis-network/standalone/umd/vis-network.min.js
(7.4.0, I believe)
Bug in solver() in HierarchicalRepulsionSolver.js:70
distance is defined as const but change is attempted in the code.
image

Thomaash:
I have no idea why it was assigned in the first place. The new value was not used for anything at all.

Closes #555.


Thomaash:
The distance is assigned and never used again. Unless I missed something in which case I'd like to know about it.

PR #592, opened 2020-03-30
Don't override Node options with Group options

### LINKED ISSUE #408: groups properties override individual
Not sure if it is as suppose to be. However it is weird that if you have a group properties and you set a node to that group and put property to different value then it is being overridden with group properties. Node borderWidth is 5 in the following example, however i would expect 1. 

        var nodes = new vis.DataSet([
            {
                id: 1,
                label:
                    [
                        "I'm an alligator,",
                        "I'm a mama-papa coming for you.",
                        "I'm the space invader,",
                        "I'll be a rock 'n' rollin' bitch for you."
                    ].join("\n"),
                x: 0,
                y: 0,
                color: {
                    background: "transparent"
                },
                group: "test",
                borderWidth: 1
            }
        ]);

        // create a network
        var data = {
            nodes: nodes,
            edges: []
        };
        var options = {
            groups: {
                test: {
                    borderWidth: 5
                }
            }
        };

Akrog:
Node options are being overridden by the values defined by the Group, which should be the case, since more specific options (node) should have priority over more generic ones (group).

This [was fixed for `color`](https://github.com/almende/vis/issues/3713), but there are still many useful options such as borderWidth, image, and shape that are being overridden.

With this PR we fix this by not merging from the group any new option that has been set for the node.

Fix #408

### DIFF
--- a/lib/network/modules/components/Node.js
+++ b/lib/network/modules/components/Node.js
@@ -247,12 +247,10 @@ class Node {
       }
     }
 
-    // Skip merging of group font options into parent; these are required to be distinct for labels
-    // Also skip mergin of color IF it is already defined in the node itself. This is to avoid the color of the
-    // group overriding the color set at the node level
-    // TODO: It might not be a good idea either to merge the rest of the options, investigate this.
-    var skipProperties = ['font'];
-    if (newOptions !== undefined && newOptions.color !== undefined && newOptions.color != null) skipProperties.push('color');
+    // Skip any new option to avoid them being overridden by the group options.
+    const skipProperties = Object.getOwnPropertyNames(newOptions).filter(p => newOptions[p] != null);
+    // Always skip merging group font options into parent; these are required to be distinct for labels
+    skipProperties.push('font');
     util.selectiveNotDeepExtend(skipProperties, parentOptions, groupObj);
 
     // the color object needs to be completely defined.


lib/network/modules/components/Node.js
```diff
@@ -219,12 +219,10 @@ class Node {
...
     var groupObj = groupList.get(group);
 
-    // Skip merging of group font options into parent; these are required to be distinct for labels
-    // Also skip mergin of color IF it is already defined in the node itself. This is to avoid the color of the
-    // group overriding the color set at the node level
-    // TODO: It might not be a good idea either to merge the rest of the options, investigate this.
-    var skipProperties = ['font'];
-    if (newOptions !== undefined && newOptions.color !== undefined && newOptions.color != null) skipProperties.push('color');
+    // Skip any new option to avoid them being overridden by the group options.
+    var skipProperties = Object.getOwnPropertyNames(newOptions).filter(p => newOptions[p] !== undefined && newOptions[p] != null);
```
Thomaash: If the first condition is true, the second will be true too. In other words the first one is useless.

Also could you use const, please? I'd like to purge var from the codebase eventually.

Akrog: Sure, thanks for the quick review.

PR #629, opened 2020-04-13
Add min/max zoom support

### LINKED ISSUE #574: New option: minZoom
Feature request:

Please add a new option `minZoom` to limit minimal zoom level. Currently, probably, there no limit to zoom out and even large graph can be "resized" until it's ~1px big, which makes no sense. I would like to set the limit so the labels are always readable.

I've tried one workaround from the internet, but it's buggy. So a native solution would be great.

gregdeane:
Closes #574


test/configure-options.test.ts
```diff
@@ -78,7 +78,11 @@ function traverseCO(root: any, rootPath: string[] = []): void {
                 typeof v === "boolean"
             )
           );
-        } else if (typeof value === "boolean" || typeof value === "string") {
+        } else if (
+          typeof value === "boolean" ||
+          typeof value === "string" ||
+          typeof value === "number"
+        ) {
```
Thomaash, line 85: What does number mean?

lib/network/options.js
```diff
@@ -667,6 +669,8 @@ let configureOptions = {
     selectConnectedEdges: true,
     hoverConnectedEdges: true,
     tooltipDelay: [300, 0, 1000, 25],
+    zoomMin: 0.00001,
+    zoomMax: 10,
```
Thomaash, line 673: This should be a range of possible values (like the value right above it). The configurator isn't able to handle single number.

test/configure-options.test.ts
gregdeane: It was to handle the non-range min/max values. I've reverted this change.

lib/network/options.js
gregdeane: Updated.

gregdeane:
@Thomaash Made the requested changes. I hope I got the ranges right.

lib/network/options.js
```diff
@@ -667,6 +669,8 @@ let configureOptions = {
     selectConnectedEdges: true,
     hoverConnectedEdges: true,
     tooltipDelay: [300, 0, 1000, 25],
+    zoomMin: [0.02, 0, 0.1, 0.005],
+    zoomMax: [10, 9.995, 10, 0.005],
```
Thomaash, line 673: I don't get this. Why is the range so small?.

test/configure-options.test.ts
Thomaash: Just to explain: This wouldn't handle anything. The purpose of this test is to catch some common mistakes (like misunderstaing the four number tuple syntax). It would still fail in runtime, you just wouldn't be warned in advance.

lib/network/options.js
gregdeane: Mainly because min/max are separate props. But I'm happy to put the range you think is appropriate. What's your suggestion for both props?

gregdeane:
> t still doesn't refuse NaN. Even though NaN works as no limit, it's a very bad idea to have NaN just floating around in unexpected places. This shouldn't be allowed to pass through setOptions.

Ah, sorry I missed that one.

lib/network/options.js
Thomaash: I'd go with 0 through 10 for both. The input validation should handle cases where max is smaller than min etc.

gregdeane: Will do, thank you.

gregdeane:
> It still doesn't refuse NaN. Even though NaN works as no limit, it's a very bad idea to have NaN just floating around in unexpected places. This shouldn't be allowed to pass through setOptions.

@Thomaash I'm having a little trouble understanding where exactly to handle the `NaN`. I see that the `Validator` catches the fact that a string should be used (for example).

In `InteractionHandler.setOptions()`, I could check if the values exist and if they are anything but numbers, use default values instead but this seems a bit hacky.

Can you explain a bit more what you want to see and point me to the code where you think it should go?

Thomaash:
> > It still doesn't refuse NaN. Even though NaN works as no limit, it's a very bad idea to have NaN just floating around in unexpected places. This shouldn't be allowed to pass through setOptions.
> 
> @Thomaash I'm having a little trouble understanding where exactly to handle the `NaN`. I see that the `Validator` catches the fact that a string should not be used (for example).
> 
> In `InteractionHandler.setOptions()`, I could check if the values exist and if they are anything but numbers, use default values instead. But this seems a bit hacky.
> 
> Can you explain a bit more what you want to see and point me to the code where you think it should go?

It isn't a bit hacky, it's very hacky.

Yeah, basically check if the value is fine and if not, just ignore it (as if it wasn't even there in the first place, don't reset it to default value simply because there's invalid input) and console error a message about it.

gregdeane:
> Yeah, basically check if the value is fine and if not, just ignore it (as if it wasn't even there in the first place, don't reset it to default value simply because there's invalid input) and console error a message about it.

In the `InteractionHandler.setOptions()` method?

Thomaash:
> > Yeah, basically check if the value is fine and if not, just ignore it (as if it wasn't even there in the first place, don't reset it to default value simply because there's invalid input) and console error a message about it.
> 
> In the `InteractionHandler.setOptions()` method? Where exactly is the "non-hacky" place to do this?

There's no non hacky place. There is some replacement in the workings, but that's stalled due to lack of time. Just trial-and-error a place in the method where it works and don't worry about it.

gregdeane:
@Thomaash Please have another look. Hope this solution fits the bill.

gregdeane:
@Thomaash Apologies for any misunderstandings. This is what I plan on doing. What do you think?

```javascript
setOptions(options) {
  if (options !== undefined) {
    // remove zoomMin and/or zoomMax if they aren't numbers
    let { zoomMin, zoomMax } = options;

    if (isNaN(zoomMin)) {
      delete options.zoomMin;
    }

    if (isNaN(zoomMax)) {
      delete options.zoomMax;
    }

    // extend all but the values in fields
    let fields = ['hideEdgesOnDrag', 'hideEdgesOnZoom', 'hideNodesOnDrag','keyboard','multiselect','selectable','selectConnectedEdges'];
    util.selectiveNotDeepExtend(fields, this.options, options);

    // merge the keyboard options in.
    util.mergeOptions(this.options, options, 'keyboard');

    if (options.tooltip) {
      util.extend(this.options.tooltip, options.tooltip);
      if (options.tooltip.color) {
        this.options.tooltip.color = util.parseColor(options.tooltip.color);
      }
    }
 }

  this.navigationHandler.setOptions(this.options);
}
```

Additionally, based on [this description](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN#Confusing_special-case_behavior), do you think we need to use `Number.isNaN() ` or the following polyfill?

```javascript
var isNaN = function(value) {
    var n = Number(value);
    return n !== n;
};
```

If you aren't satisfied with the above plan, please tell me exactly what you want.

Thanks!

Thomaash:
> @Thomaash Apologies for any misunderstandings. This is what I plan on doing. What do you think?
> 
> ```js
> setOptions(options) {
>   if (options !== undefined) {
>     // remove zoomMin and/or zoomMax if they aren't numbers
>     let { zoomMin, zoomMax } = options;
> 
>     if (isNaN(zoomMin)) {
>       delete options.zoomMin;
>     }
> 
>     if (isNaN(zoomMax)) {
>       delete options.zoomMax;
>     }
> 
>     // extend all but the values in fields
>     let fields = ['hideEdgesOnDrag', 'hideEdgesOnZoom', 'hideNodesOnDrag','keyboard','multiselect','selectable','selectConnectedEdges'];
>     util.selectiveNotDeepExtend(fields, this.options, options);
> 
>     // merge the keyboard options in.
>     util.mergeOptions(this.options, options, 'keyboard');
> 
>     if (options.tooltip) {
>       util.extend(this.options.tooltip, options.tooltip);
>       if (options.tooltip.color) {
>         this.options.tooltip.color = util.parseColor(options.tooltip.color);
>       }
>     }
>  }
> 
>   this.navigationHandler.setOptions(this.options);
> }
> ```
> 
> Additionally, based on [this description](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN#Confusing_special-case_behavior), do you think we need to use `Number.isNaN() ` or the following polyfill?
> 
> ```js
> var isNaN = function(value) {
>     var n = Number(value);
>     return n !== n;
> };
> ```
> 
> If you aren't satisfied with the above plan, please tell me exactly what you want.
> 
> Thanks!

Polyfilling is done automatically, you can just use `Number.isNaN()`, `isNaN()` is just weird.

It's better to test for valid values (you can always negate it with `!(…)`) rather than invalid ones. The way you do it is very prone to forgetting some invalid values. For example what does -7 mean for zoom? It's valid according to your code.

Also use `const` if you don't reassign it later on.

gregdeane:
@Thomaash I made the updates. I also added a specific check for `NaN` to make it as robust as possible.

BTW, the `examples` tests are failing. I don't think it's related to my changes though.

lib/network/modules/InteractionHandler.js
```diff
@@ -77,21 +77,21 @@ class InteractionHandler {
...
-      if (typeof zoomMin !== undefined && typeof zoomMin !== 'number') {
-        delete this.options.zoomMin;
+      if (!util.isNumber(zoomMin) || Number.isNaN(zoomMin)) {
+        delete options.zoomMin;
       }
 
-      if (typeof zoomMax !== undefined && typeof zoomMax !== 'number') {
-        delete this.options.zoomMax;
+      if (!util.isNumber(zoomMax) || Number.isNaN(zoomMax)) {
+        delete options.zoomMax;
```
Thomaash: What if the maximum is smaller than the minimum? What if someone inputs negative number? These situations don't seem okay to me.

gregdeane:
@Thomaash Updated.

lib/network/modules/InteractionHandler.js
```diff
@@ -705,6 +708,28 @@ class InteractionHandler {
...
+  /**
+   * Ensure zoom min/max values are acceptable
+   * If they aren't, remove them from `options`
+   * @private
+   */
+  _checkRemoveZoomMinMax(options) {
+    const { zoomMin, zoomMax } = options;
+
+    const notValid = value => (
+      !util.isNumber(value) || Number.isNaN(value) || value < 0
```
Thomaash, line 722: Why not test for:
```javascript
!(util.isNumber(value) && value >= 0)
```
It's simpler, more readable and covers everything (test all the valid cases, there's generally way fewer of them than the invalid ones and considerable less risk that you'll forget something).

Also is zero zoom really okay? It wasn't allowed before.

lib/network/modules/InteractionHandler.js
```diff
@@ -705,6 +708,28 @@ class InteractionHandler {
...
+    );
+
+    if (notValid(zoomMin) || zoomMin > zoomMax) {
```
Thomaash, line 725: Why can't they both be the same? Wouldn't that just pin the zoom level to a single value? What's wrong with that?

gregdeane: I'm not clear on what you mean here. Do you mean that the OR condition isn't in the right place (e.g. you want `notValid` to also check for the `zoomMin > zoomMax` condition)?

Taking into consideration your other comment above, this is what I have now. Can you explain a bit more about the "Why can't they both be the same?" comment?

```javascript
this.defaultOptions = {
  ...
  zoomMin: 0.00001,
  zoomMax: 10
  ...
};

const isValid = value => (
  util.isNumber(value)
  && value > this.defaultOptions.zoomMin
  && value < this.defaultOptions.zoomMax
);

if (!isValid(zoomMin) || zoomMin > zoomMax) {
  delete options.zoomMin;
}

if (!isValid(zoomMax) || zoomMax < zoomMin) {
  delete options.zoomMax;
}
```

Thomaash: Definitely not. I mean why can't min and max zoom be the same.

gregdeane: Ah, I see. Okay, will factor that in.

gregdeane:
@Thomaash The code now allows the min/max values to equal the default min/max. Additionally, both min/max values can equal each other and the zoom will be pinned to one value.

gregdeane:
@Thomaash I'm being very detailed below but I'm not suggesting you don't already know this. I am just clarifying.

> I really don't get this. Why did you limit them by default values?

They are limited by defaults in the [current implementation](https://github.com/visjs/vis-network/blob/master/lib/network/modules/InteractionHandler.js#L456). I'm just following the same logic. 

> Also according to your code they can't be the same, you test with zoomMin > zoomMax and zoomMax < zoomMin which prevents it

The test that allows the values to be the same is here:

```javascript
const isValid = value => (
  util.isNumber(value)
  && value >= this.defaultOptions.zoomMin
  && value <= this.defaultOptions.zoomMax
);
```

The code above checks only that the min/max values are valid, not whether they are the same. Thus, if both values are set to `5`, the two values are considered valid and will pass the test.

Then, the next check is performed below:

```javascript
if (!isValid(zoomMin) || zoomMin > zoomMax) {
  delete options.zoomMin;
}

if (!isValid(zoomMax) || zoomMax < zoomMin) {
  delete options.zoomMax;
}
```

In the above code, assuming `isValid` was `true`, we check only if the value is **greater than** or **less than**, not whether they are equal. If they are equal, they won't be greater than or less than, thus the condition will be false.

This is proven by the following:

```javascript
// interactionEvent.html
var options = {
  interaction: {
    ...
    zoomMin: 5,
    zoomMax: 5,
    ...
  }
};

// `this.options` after `util.mergeOptions()` is run
...
zoomMin: 5,
zoomMax: 5,
...
```

lib/network/modules/InteractionHandler.js
```diff
@@ -705,6 +710,30 @@ class InteractionHandler {
...
+    const { zoomMin, zoomMax } = options;
+
+    const isValid = value => (
+      util.isNumber(value)
+        && value >= this.defaultOptions.zoomMin
+        && value <= this.defaultOptions.zoomMax
+    );
+
+    if (!isValid(zoomMin) || zoomMin > zoomMax) {
+      delete options.zoomMin;
```
Thomaash, line 730: It would be useful to say what's wrong. That is `console.error("There's X wrong with zoomMin.")`.

lib/network/modules/InteractionHandler.js
```diff
@@ -705,6 +710,30 @@ class InteractionHandler {
...
+    }
+
+    if (!isValid(zoomMax) || zoomMax < zoomMin) {
+      delete options.zoomMax;
```
Thomaash, line 734: dtto

gregdeane:
@Thomaash Made the updates.

lib/network/modules/InteractionHandler.js
```diff
@@ -721,16 +719,18 @@ class InteractionHandler {
...
-      util.isNumber(value)
-        && value >= this.defaultOptions.zoomMin
-        && value <= this.defaultOptions.zoomMax
+      util.isNumber(value) && value > 0
     );
 
-    if (!isValid(zoomMin) || zoomMin > zoomMax) {
+    // don't bother unless `zoomMin` is passed as an option
+    if (zoomMin !== undefined && (!isValid(zoomMin) || zoomMin > zoomMax)) {
+      console.log('`zoomMin` is invalid or greater than `zoomMax`.')
```
Thomaash, line 727: `console.log` is for development purposes only. This is an error.

lib/network/modules/InteractionHandler.js
```diff
@@ -721,16 +719,18 @@ class InteractionHandler {
...
       delete options.zoomMin;
     }
 
-    if (!isValid(zoomMax) || zoomMax < zoomMin) {
+    // don't bother unless `zoomMax` is passed as an option
+    if (zoomMax !== undefined && (!isValid(zoomMax) || zoomMax < zoomMin)) {
+      console.log('`zoomMax` is invalid or less than `zoomMin`.')
```
Thomaash, line 733: dtto

gregdeane:
@Thomaash Made the updates.

gregdeane:
Just checking in on this. Anyway we can merge?

gregdeane:
@Thomaash Can we merge this?

yotamberk:
@Thomaash Can you re-review this?
it seems like the changes have been done

gregdeane:
> Sorry for the delay, can't find any time for this project lately. It looks good except for one (hopefully) last thing: If I set the zoom limits it's not enforced right away. The user has to zoom for the limit to take action. For example if min zoom is 5 the network shouldn't open with zoom 1.

@Thomaash I've having a bit of trouble finding the spot where this change should be made. I've narrowed it down to [here](https://github.com/visjs/vis-network/blob/master/lib/network/modules/Canvas.js#L63) but I'm not convinced this is the correct place. At this point, `options` contains the `interaction` prop (if set).

My thought was that if I am able to merge `zoomMin` into the `options` object at this point, then I can use it to set `scale` [here](https://github.com/visjs/vis-network/blob/master/lib/network/modules/Canvas.js#L145).

It would be of great help if you could point me in the right direction and perhaps even suggest how I might implement the change.

Thanks!

Thomaash:
Ideally all changes to scale would be handled in one place (View.js seems like a good candidate) and receive requests from other places (e.g. `this.body.emitter.emit("scale", { to: 4.7 })` or `this.body.emitter.emit("scale", { by: -0.3, at: { x: 77, y: 234 } })`). Then there wouldn't be the problem that the same configuration and logic has to be in multiple places.

I don't know how eager you are to dive into this. However I don't see much of an alternative here. You could keep it fragmented as it is now but then you would still need to establish some form of communication because there's also fit, resizing and maybe some other scale altering stuff I can't remember right now.

gregdeane:
> Ideally all changes to scale would be handled in one place (View.js seems like a good candidate) and receive requests from other places (e.g. `this.body.emitter.emit("scale", { to: 4.7 })` or `this.body.emitter.emit("scale", { by: -0.3, at: { x: 77, y: 234 } })`). Then there wouldn't be the problem that the same configuration and logic has to be in multiple places.
> 
> I don't know how eager you are to dive into this. However I don't see much of an alternative here. You could keep it fragmented as it is now but then you would still need to establish some form of communication because there's also fit, resizing and maybe some other scale altering stuff I can't remember right now.

Just want to let you know that I haven't forgotten about this PR. Work has been insanely busy lately and will likely continue to be for another week or two. I'll jump back on this PR as soon as I can.

gregdeane:
@Thomaash Unfortunately, my schedule won't be cleared up for quite some time. If there's a break, I'll get back on this feature. Otherwise, I'm out for the foreseeable future. This will most likely be the case.

Please understand that the following is detailed not with an accusatory tone but with a friendly tone. No need to take offense.

I suggest that for future PR's, everything you wish to see is commented upon during the first review. You want things a very specific way and that's fine. But it would have been beneficial to let me know in the very beginning exactly what you wanted. The back-and-forth went on much too long in my opinion and made it more difficult to continue working on this feature.

You may lay blame on me for not having the code properly written or up to vis-network standards. Again, fine. Just explain what you expect during the first review. If you wish to have something written a specific way, say as much. Just provide an example (e.g. code snippet) of what you want.

It should not take so long to contribute a small change, in my opinion.

Anyway, I hope to find time to address this topic again. But if not, I hope someone else will finish it up.

Thomaash:
Thanks for the feedback. It makes a lot of sense and I'll try to give better and more detailed reviews right away. Anyway it's fine if you can't find the time to work on this, I can understand that.

gregdeane:
@Thomaash And thank you for your understanding and consideration on the feedback. Much appreciated!

Jacse:
What's the status of this? @Thomaash are you taking over or do you need someone else to? I'm guessing you'll want consistency with #1134 which you just merged.

PR #637, opened 2020-04-15
Issue #634 - getPositions signature fix, getPosition feature.

### LINKED ISSUE #634: Network.getPositions() TS type/documentation does not align with actual implementation.
Hello there!

One of our projects was working on hooking into the afterDrawing event and drawing some more objects around a given node. In doing this, we were attempting to use Network.getPositions in the manner defined by the ts definitions/documentation:
```js
  /**
   * Returns the x y positions in canvas space of the nodes with the supplied nodeIds as an object.
   *
   * Alternative inputs are a String containing a nodeId or nothing.
   * When a String is supplied, the position of the node corresponding to the ID is returned.
   * When nothing is supplied, the positions of all nodes are returned.
   */
  getPositions(nodeIds?: IdType[]): { [nodeId: string]: Position };
  getPositions(nodeId: IdType): Position;
```

The method getPositions(nodeId): Position is not actually implemented as defined: https://github.com/visjs/vis-network/blob/93a79d73099a8f13088ea93c6ad0210616b1a177/lib/network/modules/NodesHandler.js#L403-L428

So, either the ts definition for that overload should be removed/first one modified to accept the string only and return keyed position, or the getPositions() method on line 417 should be altered to return the x/y position object directly, rather than the dataArray object.

Or a new method getPosition(nodeId: IdType): Position rather than the overload and split return type?

I'm happy to open a PR with whatever direction you guys want to go.

tzfx:
This pull request closes #634 

- Typescript definition for `getPositions` has been modified to reflect its implementation.
- A new method, `getPosition(id: IdType)` has been created, returning only Position data.
- Documentation and tests have been added for this new method.
- Documentation for `getPositions` has been updated with more examples showing the expected behaviour, as well as a cautionary note on its silent failing behaviour.

For this PR I chose to leave the implementation of `getPositions` as-is and only update the typescript definition. The implementation of `getPosition` includes the throwing of some errors, namely `TypeError`, in the case of no arguments supplied, and `ReferenceError` in the case of a id supplied that was missing from the underlying dataset.

This differs from `getPositions`, which either fails silently or strips out missing ids from the return.

Let me know if I missed anything as far as test suites/documentation goes-- I had trouble verifying the cypress suite and the end-to-end stuff on my computer, but all the unit tests ran without issue.


docs/network/index.html
```diff
@@ -854,25 +854,62 @@ <h2 id="methods">Methods</h2>
                 <td colspan="2">Methods to get information on nodes and edges.</td>
             </tr>
             <tr class="collapsible toggle" onclick="toggleTable('methodTable','getPositions', this);">
-                <td colspan="2"><span parent="getPositions" class="right-caret" id="method_getPositions"></span> getPositions(<code><i>[Array of
-                    nodeIds]</i></code>)
+                <td colspan="2"><span parent="getPositions" class="right-caret" id="method_getPositions"></span>
+                    getPositions(<code><i>[Array of nodeIds]</i></code>, <code><i>[nodeId]</i></code>)
```
Thomaash, line 858: This looks like two optional parameters, ie:
```typescript
getPositions(nodes: NodeId[], nodeId: NodeId): unknown
```

docs/network/index.html
```diff
@@ -854,25 +854,62 @@ <h2 id="methods">Methods</h2>
...
+
+
+// A single node.
+network.getPositions('a123');
+>   {
+        a123: { x: 5, y: 12 }
+    }
+                    
+                    </pre>
                     Alternative inputs are a String containing a nodeId or nothing. When a String is supplied, the
```
Thomaash, line 890: I have no idea why the word string was with capital s, it makes no sense to me. It should be string with lower case s.

Similar issue is with id/ID. We generally treat it as a standard word in code (nodeId, not nodeID) but it's pretty inconsistent in the docs. We should probably stick with id for the sake of consistency with the code (it's short for identifier, which is a single word, though the common spelling is different).

types/network/Network.d.ts
```diff
@@ -350,8 +350,15 @@ export class Network {
...
    */
-  getPositions(nodeIds?: IdType[]): { [nodeId: string]: Position };
-  getPositions(nodeId: IdType): Position;
+  getPositions(nodeIds?: IdType[] | IdType): { [nodeId: string]: Position };
+  
+  /**
+   * Retrieves the x y position of a specific id.
+   * @param {IdType} id The id to retrieve.
+   * @throws {TypeError} If no id is included.
+   * @throws {ReferenceError} If an invalid id is provided.
```
Thomaash, line 359: Don't use JSDoc syntax, use TSDoc instead. Param types are already in the code, so:
```
* @param id - The id to retrieve.
```
is enough. Also throws syntax is following (more details in https://github.com/microsoft/tsdoc/pull/175):
```
* @throws {@link TypeError}
* If no id is included.
```
Leave an empty line between the text and the params. And one more thing, your descriptions are quite misleading. For example the param doesn't retrieve the id, as you description suggests, it retrieves the position of the node with given id. Reference error is not thrown if an invalid id is provided but if no node with such id, which may be completely valid, is in the network. The phrase “If no id is included” is also quite weird (included where? how? etc.)

docs/network/index.html
tzfx: Totally right! Should be fixed with the latest push.

docs/network/index.html
tzfx: Yeaaaah- I tried to look through the documentation to try to figure out a standard, but didn't really see one. Moved to string/id for the documentation that I added/altered.

types/network/Network.d.ts
```diff
@@ -344,14 +344,38 @@ export class Network {
...
+   * Returns the x y positions in canvas space of all nodes
    *
-   * Alternative inputs are a String containing a nodeId or nothing.
-   * When a String is supplied, the position of the node corresponding to the ID is returned.
-   * When nothing is supplied, the positions of all nodes are returned.
+   * @returns A an object containing the x y positions in canvas space of all nodes in the network, keyed by id.
    */
-  getPositions(nodeIds?: IdType[]): { [nodeId: string]: Position };
-  getPositions(nodeId: IdType): Position;
+  getPositions(): { [nodeId: string]: Position };
```
tzfx, line 351: I broke up getPositions() and getPositions(nodeIds: IdType[] | IdType) for my own sanity in documentation.

Thomaash: Why? You could simply say that `nodeIds` defaults to ids of all the nodes currently in the network.

Also how can it return nothing? Doesn't it always return at least an empty object?

tzfx: Sure, I can do that.

You are correct on the nothing- I've revised the docs to indicate the an empty object will be returned. Next push should be correct in this regard.

PR #761, opened 2020-05-28
Add possibility to move endpoints.

pahrg:
**New feature**
Sometimes you want to have the possibility to adjust the endpoints, for example when you want your arrows be placed on the nodes border.

Separate line resp. endpoint rendering plus add property endpointsAdjustedPosition.
Take care of the dependent of the arrowStrikethrough.
Add documentation.

83020098-4a113880-a028-11ea-9e19-7309f5ed6d8a


docs/network/edges.html
```diff
@@ -324,6 +325,12 @@ <h3>Options</h3>
             <td><code>Object</code></td>
             <td>Exactly the same as the to object but with an arrowhead at the from node of the edge.</td>
         </tr>
+        <tr>
+            <td class="indent">endpointsAdjustedPosition</td>
+            <td>Number</td>
+            <td><code>0</code></td>
+            <td>Set a number to adjust the endpoints. It can be useful if you want to have the endpoints placed on the nodes border. Negative numbers move the endpoints nearer the nodes midpoint, and positive numbers does the opposite. It's only possible when the arrowStrikethrough is true.</td>
+        </tr>
```
Thomaash, line 333: We have Korean docs. Could you add it there too, please?

PS: You can replace Korean by TODO if you don't speak Korean. Someone will hopefully translate it later.

lib/network/modules/components/Edge.js
```diff
@@ -543,6 +545,26 @@ class Edge {
       return;
     }
 
+    // get the via node from the edge type
+    let viaNode = this.edgeType.getViaNode();
```
Thomaash, line 549: If it can be const, make it const, thanks.

lib/network/modules/components/Node.js
```diff
@@ -716,6 +716,18 @@ class Node {
...
     }
   }
+
+  /**
+  * Find out the endpointsAdjustedPosition value
+  * @private
+  */
+  _getEndpointsAdjustedPosition() {
+    if (this.edges.length > 0) {
+      return this.edges[0].globalOptions.arrowStrikethrough === true ? this.edges[0].globalOptions.endpointsAdjustedPosition : 0;
```
Thomaash, line 726: What does this do? Are you taking the global edge options from the first edge? What sense does this make?

lib/network/options.js
```diff
@@ -68,6 +68,7 @@ let allOptions = {
       },
       __type__: { string: ["from", "to", "middle"], object }
     },
+    endpointsAdjustedPosition: { number },
```
Thomaash, line 71: I don't like the name. Ideally you should see the name and immediately know what it does. When I saw this for the first time I thought it overrides the old positioning, only after I read the docs part I understood (I skimmed through the code first, then read the PR text). I suggest something like `endPointOffset`. Of course your free to suggest alternatives or improvements.

pahrg:
Thank's for your attention
We are glad for some discussion with you.

docs/network/edges.html
pahrg: Sorry, I missed that. Of course I will add TODO notes.

lib/network/modules/components/Edge.js
pahrg: NP, I will do.

lib/network/options.js
pahrg: I think endPointOffset is a good name, thanks.

lib/network/modules/components/Node.js
pahrg: I guess that the property should be placed in the edges options in connection with the arrows properties. 
Maybe I missed it but is there a better solving to grab the value?

Thomaash: We have complex system of option merging. That is global options are automatically merged into each edge if it doesn't have given option of its own. There should never be the need to touch global options directly and I don't see a reason why this should be global only. Ideally this would be configurable per edge end point (something like `endPointOffset: number | { from?: number, to?: number }`).

Per node configuration is also worth discussing. Because different nodes may require different offsets due to different styling. Although different end points may also need different offsets even when connected to the same node. Per edge end point configuration would be most flexible, I think.

pahrg: Yes you are right, per edge and end point should be the best configuration. 
Thanks a lot :)

lib/network/modules/CanvasRenderer.js
```diff
@@ -411,6 +423,24 @@ class CanvasRenderer {
...
+   * @param {CanvasRenderingContext2D} ctx  2D context of a HTML canvas
+   * @private
+   */
+  _drawArrows(ctx) {
+    let edges = this.body.edges;
+    let edgeIndices = this.body.edgeIndices;
+    let edge;
+
+    for (let i = 0; i < edgeIndices.length; i++) {
+      edge = edges[edgeIndices[i]];
```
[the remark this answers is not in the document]
pahrg: I will change it.

pahrg:
The already implemented validation take care of the input of the new properties. For example if you put some text in the number fields the Validator.js response with invalid type error. 
The endpoints can actually just be drawn somewhere on the edge, the midpoints of the nodes will be the limits. I don't think we need any specific validation for min/max values.

Thomaash:
> The already implemented validation take care of the input of the new properties. For example if you put some text in the number fields the Validator.js response with invalid type error.
> The endpoints can actually just be drawn somewhere on the edge, the midpoints of the nodes will be the limits. I don't think we need any specific validation for min/max values.

Yes, we do need validation. The validator only checks types, but that something is number doesn't mean it's going to work. Will it work if you put there `Number.NaN`, `Number.MAX_VALUE` or `Number.NEGATIVE_INFINITY`? The last time I tried the endpoints simply disappeared and I don't think this is acceptable to end users.

pahrg:
Yes, I see. If we put in max values or positive/negative infinity numbers the endpoints will be placed on one of the nodes midpoint depending of where the property (from or to) is defined. The same will happen with NaN.
But if we add validation for the input values how will we handle it, should we respond with some error like the type validation? What do you think?

Thomaash:
Just log it with `console.error` and ignore the invalid value.

lib/network/modules/components/Edge.js
```diff
@@ -144,6 +144,23 @@ class Edge {
     // only deep extend the items in the field array. These do not have shorthand.
     util.selectiveDeepExtend(fields, parentOptions, newOptions, allowDeletion);
 
+    // Only use endPointOffset values (from and to) if it's valid values
+    if (newOptions.endPointOffset !== undefined) {
```
Thomaash, line 148: This shouldn't report values that are not there as invalid, i.e. `endPointOffset: { from: 7 }` shouldn't be reported.

lib/network/modules/components/Edge.js
```diff
@@ -144,6 +144,23 @@ class Edge {
...
+      if (ComponentUtil.isValidInteger(newOptions.endPointOffset.from)) {
```
Thomaash, line 149: This is just a question at this point, not a request for change: Why are only integers valid? Would something terrible happen if it was for example 1.2?

lib/network/modules/components/Edge.js
```diff
@@ -144,6 +144,23 @@ class Edge {
...
+        parentOptions.endPointOffset.from = newOptions.endPointOffset.from;
+      } else {
+        parentOptions.endPointOffset.from = globalOptions.endPointOffset.from !== undefined ? globalOptions.endPointOffset.from : 0;
+        console.log('endPointOffset.from is not a valid number');
```
Thomaash, line 153: Use `console.error()`, please.

lib/network/modules/components/Edge.js
```diff
@@ -144,6 +144,23 @@ class Edge {
...
+      }
+
+      if (ComponentUtil.isValidInteger(newOptions.endPointOffset.to)) { 
+        parentOptions.endPointOffset.to = newOptions.endPointOffset.to;
+      } else {
+        parentOptions.endPointOffset.to = globalOptions.endPointOffset.to !== undefined ? globalOptions.endPointOffset.to : 0;
+        console.log('endPointOffset.to is not a valid number')
```
Thomaash, line 160: dtto

lib/network/modules/components/Edge.js
pahrg: Yes of course.

lib/network/modules/components/Edge.js
pahrg: I have just test with integers but it should not matters. Do you think there is a need for floating numbers? Maybe if the place of the endpoints have to be very exactly..?

pahrg:
You are welcome

That's strange... if I supply Number.NaN I get an error that says that isn't a valid number. The edge will be drawed and the failing endpoint will use the globaloption value if defined or the value is set to 0.

lib/network/modules/components/Edge.js
pahrg: You have to define both from and to values.

pahrg: I change it so that you can define just one endpoint on a edge.

lib/network/modules/components/Edge.js
pahrg: I change so also floats is valid.

lib/network/modules/components/shared/ComponentUtil.js
```diff
@@ -116,13 +116,13 @@ class ComponentUtil {
...
+   * Check if given value is acceptable as an number.
    *
    * @param {*} number value to check; can be anything at this point
-   * @returns {boolean} true if valid integer value, false otherwise
+   * @returns {boolean} true if valid number value, false otherwise
    */
-  static isValidInteger(number) {
-    return (typeof number === 'number' && Number.isInteger(number) && Number.isFinite(number));
+  static isValidNumber(number) {
+    return (typeof number === 'number' && Number.isFinite(number));
```
Thomaash, line 125: Doesn't `isValidNumber(x) === Number.isFinite(x)` for any `x`?

Anyway “acceptable as an number” [sic] is very vague.

pahrg: As I see it will isFinite not let through any numbers who not are acceptable as a finite number.
Maybe it's to weak validation, have you any suggestions or ideas?

Thomaash: I don't see a reason why it would be too weak. Personally I would just ditch `isValidInteger` and use `Number.isFinite` directly.

pahrg:
Thank you very much :)

PR #788, opened 2020-06-06
chore: automerge non-major dependencies versions

yotamberk:
I've been running after merging patches and minor dependencies from npm.
Thanks to renovate, this is does open PRs automatically but it does seem to spam versions and PRs which kinda sucks sometimes because they should be simply automatically merged in...

This PR will automatically merge them with no human interventions.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,5 +1,11 @@
 {
   "extends": [
     "config:base"
+  ],
+  "packageRules": [
+    {
+      "updateTypes": ["minor", "patch", "pin", "digest"],
+      "automerge": true
+    }
   ]
 }


Thomaash:
Hi @yotamberk,

I'm not sure this is going to work. Won't it wait for manual reviews as per branch protection?

yotamberk:
Nope. It says in the docs (https://docs.renovatebot.com/configuration-options/#automerge) that before merging, it checks that the pipeline passes

Thomaash:
No. I mean that we have branch protection on GitHub configured to prevent merging without a review. I think we'll have to turn it off for this to work because otherwise it won't be able to merge without one of us approving the PR.

yotamberk:
Ummm okay that make sense. 
I think we can allow `renovate` aaccess to merge to a specific protected branch by user

PR #817, opened 2020-06-20
feat: Multi-select nodes with shiftKey

yotamberk:
I found this feature very much needed for my projects.
multi-select


yotamberk:
My boo boo... Fixed!
@Thomaash can you check again?

lib/network/modules/InteractionHandler.js
```diff
@@ -413,19 +436,40 @@ class InteractionHandler {
...
+      this.body.selectionBox.show = false;
+      const selectionBoxPosition = this.body.selectionBox.position 
+      const selectionBoxPositionMinMax = {
+        minX: Math.min(selectionBoxPosition.start.x, selectionBoxPosition.end.x),
+        minY: Math.min(selectionBoxPosition.start.y, selectionBoxPosition.end.y),
+        maxX: Math.max(selectionBoxPosition.start.x, selectionBoxPosition.end.x),
+        maxY: Math.max(selectionBoxPosition.start.y, selectionBoxPosition.end.y),
+      }
+
+      const toBeSelectedNodes = Object.values(this.body.nodes).filter((node) => (
```
Thomaash, line 450: Shouldn't you use `this.body.nodeIndices`? Because `Object.values(this.body.nodes)` includes hidden and support nodes.

yotamberk: You're right.
I've fixed it to your suggestion so the selection does not include hidden and support nodes.

yotamberk:
The stokeColor is determined as a global color that switches all the time depending on what the graph is currently drawing. I've fixed the issue so the color is consistent to the default color.

In terms of the other points you've raised (fantastic points by the way), I think they are all important but shouldn;t be part of this PR.
1. I think deselection is an opinionated issue. I think there are many different ways to implement this and we might consider opening an issue to decide what the community thinks would be the best option. Valid options might be: 
- command + shift
- shift + drag on an already selected node
- alt + shift
- etc

2. touch screen support might be implemented by overriding the single drag, and then moving the canvas would work with 2 fingers panning (I think this is how it works on some other apps I've tried out) - this is a breaking change and I don't know if this is the best solution. 
I'm afraid that this too is an opinionated feature that I don;t have a concrete answer or solution. We can also open an issue on this.

3. configurating the color of the selection box is a valid feature request. Easy to do in the future if the community seeks it.

In summary, I think the PR is ready for merging now and the rest of the points can be done in future PRs.

Issue #713, opened 2020-05-16
[Feature Request]


cristianocca:
Shapes are great, but could we add a little more customization to them? Right now they can only be resized with the `value` property. It would be great to be able to adjust width/height of boxes, radius of circles, etc.  What's the rationale for not supporting this/

I know this can be achieved with SVG images, but it gets much heavier and harder to maintain than just passing props to the `box` shape.

Thoughts?

Thomaash:
Box is shaped according to the text inside. If the text is inside it would be pretty difficult to deal with this and also there is the question what to do with the text (downscale/upscale/overflow/etc).

As for the others, I don't really see a reason why not. It shouldn't be that difficult to achieve. However there are much more pressing issues requiring my very limited time so if you want this functionality you'll have to code it yourself (and thank you very much if you do) or find someone willing to work on it.

cristianocca:
For the `box` shape, there's already logic/options for max width/height a box is allowed to grow when there's text in it. I'm assuming this logic already deals with text wrapping and such, wouldn't it be as simple as just "fixing" the width/height of the box and let the text wrap in it?  Sorry if I'm over simplifying it, I haven't really looked into the code.

Since I can right now work around this with SVG shapes, it is not so urgent for me neither, but I would be willing to implement it if required.  Some hints about where to look at in the code would also be helpful :)

PR #835, opened 2020-06-28
Add custom shapes renderer

yotamberk:
I've noticed that custom shapes are very much needed and the fact that there are only specific types of shapes is quite limiting. (related to #713)

I've added a 'custom' shape type for nodes where you can supply a `ctxRenderer(ctx, x, y, values)` function prop which allows you access the node drawer via the `ctx` so we allow the developer to add his own shapes with canvas drawing methods. 

I've added an example in the  `\examples\network\nodeStyles\shapes.html` example.
docs will be updated soon too.

image

image


lib/network/modules/components/nodes/shapes/CustomShape.js
```diff
@@ -0,0 +1,47 @@
+'use strict';
+
+import ShapeBase from '../util/ShapeBase'
+
+/**
+ * A Hexagon Node/Cluster shape.
+ *
+ * @extends ShapeBase
+ */
+class Hexagon extends ShapeBase {
```
Thomaash, line 10:
```javascript
class CustomShape {}
```

lib/network/modules/components/nodes/util/ShapeBase.js
```diff
@@ -43,16 +43,21 @@ class ShapeBase extends NodeBase {
...
+  _drawShape(ctx, shape, sizeMultiplier, x, y, selected, hover, values, customRenderer) {
     this.resize(ctx, selected, hover, values);
     this.left = x - this.width / 2;
     this.top = y - this.height / 2;
 
     this.initContextForDraw(ctx, values);
-    getShape(shape)(ctx, x, y, values.size);
-    this.performFill(ctx, values);
+    if (shape === 'custom') {
+      customRenderer(ctx, x, y, selected, hover, values)
```
Thomaash, line 56: We should probably `ctx.save()` and `ctx.restore()` around this. Who knows what they'll set and won't clear in this function.

yotamberk: Agreed! good catch.

lib/network/modules/components/nodes/shapes/CustomShape.js
yotamberk: oops...

examples/network/nodeStyles/shapes.html
```diff
@@ -47,7 +47,33 @@
         {id: 32, font:{size:30}, size:40, label: 'big star',    shape: 'star'},
         {id: 33, font:{size:30}, size:40, label: 'big hexagon', shape: 'hexagon'},
         {id: 34, font:{size:30},          label: 'icon square', shape: 'square', icon: {code:'\uf164'} },
-        {id: 35, font:{size:30},          label: 'icon dot',    shape: 'dot',    icon: {code:'\uf165'} },        
+        {id: 35, font:{size:30},          label: 'icon dot',    shape: 'dot',    icon: {code:'\uf165'} },   
+        {
+          id: 36, 
+          label: 'custom', 
+          shape: 'custom', 
+          ctxRenderer: (ctx, x, y, selected, hover, values) => {
```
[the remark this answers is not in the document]
yotamberk: Good idea. Will do.

yotamberk:
Working on it now 😄 I'll get it in today

yotamberk:
Added the docs.

lib/network/modules/components/nodes/util/ShapeBase.js
```diff
@@ -43,16 +43,23 @@ class ShapeBase extends NodeBase {
...
+   * @param {function} customRenderer - a custom shape renderer similar to getShape(shape) functions
    * @private
    */
-  _drawShape(ctx, shape, sizeMultiplier, x, y, selected, hover, values) {
+  _drawShape(ctx, shape, sizeMultiplier, x, y, selected, hover, values, customRenderer) {
     this.resize(ctx, selected, hover, values);
     this.left = x - this.width / 2;
     this.top = y - this.height / 2;
 
     this.initContextForDraw(ctx, values);
```
Thomaash, line 54: I wouldn't run this for custom shapes. I think this is tightly coupled to built-in shapes and could cause problems later on (we couldn't change anything here as it would break people's custom shapes).

lib/network/modules/components/nodes/util/ShapeBase.js
```diff
@@ -43,16 +43,23 @@ class ShapeBase extends NodeBase {
...
-    getShape(shape)(ctx, x, y, values.size);
-    this.performFill(ctx, values);
+    if (shape === 'custom') {
+      ctx.save();
+      customRenderer({ ctx, x, y, selected, hover, values });
```
Thomaash, line 57: One more thing: I would flatten the arguments here (`{ ctx, x, y, selected, hover, ...values }`). Values is very vague and from API point of view, I don't see any reason for its existence. It would be just confusing to users that something can be directly destructured and something is hidden under values. Also someone could change the values and mess with something down the line as it's passed by reference (I'm not sure if this is the case at the moment but at some point it could be, better safe than sorry).

Alternatively it would make sense to group it in some logical way, e.g.:
```typescript
interface CustomShapeArgs {
  ctx: CanvasRenderingContext2D;
  position: { x: number; y: number };
  state: { selected: boolean; hovered: boolean };
  style: { opacity: number; color: string; ... };
}
```

lib/network/modules/components/nodes/util/ShapeBase.js
yotamberk: I'm not sure what you mean. Do you think there should be a different function for the custom shape draw?

lib/network/modules/components/nodes/util/ShapeBase.js
yotamberk: ok, makes sense.

lib/network/modules/components/nodes/util/ShapeBase.js
Thomaash: No, I mean specifically the call to `initContextForDraw`.

yotamberk: Ok so I think the solution is to move `initContextForDraw` to line 61 in the case the shape is not 'custom'

Thomaash: Yeah, maybe it would've been clearer if I said something like “move `initContextForDraw` to the else block”.

jk-12345:
Hi,

This new feature is really nice, thanks for it.

Meanwhile, we have only the size option for the node interacting zone itself, which gives us a scare of size*size. But, many rendering we could use are of a "large" or "tall" shape so that a large part of a scary interacting zone remain empty yet interacting (select, ends of edge...).

Can you add more options to this new "Custom shape" pls ? 

A rectangular interacting zone could be a great thing, for instance passing a object such as : 
size : {
    width: number,
    height: number
}

What do you think ?

Thomaash:
> Hi,
> 
> This new feature is really nice, thanks for it.
> 
> Meanwhile, we have only the size option for the node interacting zone itself, which gives us a scare of size*size. But, many rendering we could use are of a "large" or "tall" shape so that a large surface of a scary interacting zone in that case remains empty yet interacting (select, ends of edge...).
> 
> Can you add more options to this new "Custom shape" pls ?
> 
> A rectangular interacting zone could be a great thing, for instance passing an object such as :
> size : {
> width: number,
> height: number
> }
> 
> What do you think ?

I think you should open an issue for this, but otherwise yeah, sounds useful.

PR #844, opened 2020-07-06
New coordinateOrigin prop to change image/circularImage positioning

cristianoccazinsp:
This PR allows the use of (x, y) of the node as the (0, 0) position of the image instead of using width/2 and height/2.

A new shape property was added: coordinateOrigin (default 'center'). The new option `top-left` allows the image/circle image's top-left corner to be the x,y position of the node

Closes https://github.com/visjs/vis-network/issues/811
Closes https://github.com/visjs/vis-network/pull/812

### DIFF
--- a/examples/network/nodeStyles/imagesWithCoordinateOrigin.html
+++ b/examples/network/nodeStyles/imagesWithCoordinateOrigin.html
@@ -1,0 +1,113 @@
+<!doctype html>
+<html>
+<head>
+  <title>Vis Network | Node Styles | Images with alternative coordinate origin</title>
+
+  <style type="text/css">
+    #mynetwork {
+      width: 600px;
+      height: 600px;
+      border: 1px solid lightgray;
+    }
+  </style>
+
+  <script type="text/javascript" src="../../../standalone/umd/vis-network.min.js"></script>
+
+  <script type="text/javascript">
+    var nodes = null;
+    var edges = null;
+    var network = null;
+
+    var DIR = '../img/refresh-cl/';
+    var EDGE_LENGTH_MAIN = 150;
+    var EDGE_LENGTH_SUB = 50;
+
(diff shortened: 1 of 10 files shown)


examples/network/nodeStyles/imagesWithCoordinateOrigin.html
```diff
@@ -0,0 +1,113 @@
...
+      nodes.push({id: 3, label: 'Wireless', image: DIR + 'Network-Pipe-icon.png', shape: 'image'});
+      edges.push({from: 1, to: 2, length: EDGE_LENGTH_MAIN});
+      edges.push({from: 1, to: 3, length: EDGE_LENGTH_MAIN});
+
+      for (var i = 4; i <= 7; i++) {
+        nodes.push({id: i, label: 'Computer (top-left)', image: DIR + 'Hardware-My-Computer-3-icon.png', shape: 'circularImage'});
+        edges.push({from: 2, to: i, length: EDGE_LENGTH_SUB});
+      }
+
+      nodes.push({id: 101, label: 'Printer', image: DIR + 'Hardware-Printer-Blue-icon.png', shape: 'circularImage'});
```
Thomaash, line 44: Could you add center and top-left to all of them? Or if there is some reason for not including it, it should be explained somewhere as the way it is it's just confusing.

cristianoccazinsp: You are right, I didn't add the `top-left` labels because the example was initially thought to showcase that very specific example, so the `center` label is just to compare. But you are right, it makes more sense for the label to be on both cases!

cristianoccazinsp:
What happened with the tests now?

cristianoccazinsp:
Updated the example, and the tests seem to be running fine:
Screen Shot 2020-07-06 at 22 52 31

PR #892, opened 2020-07-25
fix: add label prop to ctxRenderer

yotamberk:
When trying to add a custom shape via the `ctxRenderer` prop for nodes, sometimes the label is useful to inject in the custom shape itself.
I've added this so it's very easy to use.
Except for this addition, when adding a custom shape, the label is drawn below and this is unwanted so I returned straight after a custom shape is drawn.


Thomaash:
@yotamberk 

Rethinking this, wouldn't it be better to return further instructions from the function? Something like:
```typescript
interface CtxRendererReturnValue {
  renderLabel: boolean;
}
```

This way we would avoid breaking changes (`renderLabel` property would default to `true`) and also people, who just want to put custom shape above the label, could render the shape and have standard node label for free.

yotamberk:
makes sense. I'll fix it now and resubmit

Thomaash:
Okay, I stopped the release in CI.

yotamberk:
I can maybe add another prop simply with the `renderLabel: boolean;` that defaults to `true`

Thomaash:
Do we actually know where to render the label? Because custom shape can be anything of any size. Maybe we should just skip the label altogether after all.

yotamberk:
Rethinking this, in order of doing this, we wil need to add a prop to the nodes options. I don;t think we want to do this. The original PR is what I intended

yotamberk:
Can we revert the revert?

Thomaash:
Yeah, I agree. It will be best to skip the label and let people deal with it (i.e. custom means custom).

Thomaash:
Revert reverted :smile:

yotamberk:
Thanks 😄

yotamberk:
We just need to make sure the CI runs again so it publishes properly

Thomaash:
This should be fine. I canceled the previous runs before they got to releasing so the last one should release as if the reverts didn't even happen.

PR #895, opened 2020-07-25
fix: add ctxRenderer and shape of 'custom' to options

yotamberk:
egh... forgot to add the options... I had bad errors...


yotamberk:
@Thomaash Is the realease stuck? it seems to not release this and the previous PR

PR #932, opened 2020-08-08
fix(rendering): render external node labels above arrows

### LINKED ISSUE #883: Edge arrows appear above node labels
On my network, arrows on both and and middles of edges appear drawn a layer above any node label text, while the edges themselves are underneath the node label.

Thomaash:
This places external node labels (labels outside of the node's shape) above arrows. This is how it was before #761 (the reason is that the arrows can pierce into nodes and therefore have to be above the nodes). The way this works is that the draw function returns label drawing function and the renderer invokes the label drawing function later to draw over the arrows.

The E2E test has also been extended to test all the shapes instead of just two.

Partially solves #883 (custom shape still has to be addressed).

### DIFF
--- a/cypress/integration/visual/z-index.spec.ts
+++ b/cypress/integration/visual/z-index.spec.ts
@@ -1,83 +1,118 @@
+const EXTERNAL_LABEL_SHAPES = new Set<string>([
+  "diamond",
+  "dot",
+  "hexagon",
+  "square",
+  "star",
+  "triangle",
+  "triangleDown",
+]);
+const INTERNAL_LABEL_SHAPES = new Set<string>([
+  "box",
+  "circle",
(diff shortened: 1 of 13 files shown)


Thomaash:
> does this work also for custom shape?

That's a good point, it doesn't work for custom shape. We have to rework the API for this to work.

Thomaash:
I kinda like the idea of returning multiple functions for different layers, this would be quite easy to extend in the future without breaking changes:
```typescript
function ctxRenderer(ctx) {
  // do some math here
  return {
    renderNode: () => {
      ctx.drawSomeNode();
    },
    renderNode: () => {
      ctx.drawSomeTextOutsideOfTheNode();
    },
  };
}
```

PR #950, opened 2020-08-12
feat: add ctxRenderer return width/height to calculate node sizes properly

yotamberk:
Using a custom shape renderer allows the user to add any shape desired. BUT when the shape has a size of it's own, the nodes in the network are then dependant of these sizes and will draw edges badly. This fix will allow the ctxRenderer function (the user will have to be responsible to return the `{ width, height }` object) to determine the size of his nodes and make the network function properly with these sizes.


yotamberk:
@Thomaash I see I have conflicts with the master branch. I don;t quite know how to implement my change on the master branch. Can you help me with this?

lib/network/modules/components/nodes/util/ShapeBase.js
```diff
@@ -23,13 +23,18 @@ class ShapeBase extends NodeBase {
...
+    values = { size: this.options.size }
+  ) {
     if (this.needsRefresh(selected, hover)) {
       this.labelModule.getTextSize(ctx, selected, hover);
       const size = 2 * values.size;
-      this.width = size;
-      this.height = size;
-      this.radius = 0.5*this.width;
+      this.width = this.customSizeWidth || size;
+      this.height = this.customSizeHeight || size;
```
Thomaash, line 36: Doesn't `??` make more sense here?

lib/network/modules/components/nodes/util/ShapeBase.js
```diff
@@ -97,15 +104,24 @@ class ShapeBase extends NodeBase {
...
    */
   updateBoundingBox(x, y) {
-    this.boundingBox.top = y - this.options.size;
-    this.boundingBox.left = x - this.options.size;
-    this.boundingBox.right = x + this.options.size;
-    this.boundingBox.bottom = y + this.options.size;
+    this.boundingBox.top = y - (this.customSizeHeight || this.height);
+    this.boundingBox.left = x - (this.customSizeWidth || this.width);
+    this.boundingBox.right = x + (this.customSizeWidth || this.width);
+    this.boundingBox.bottom = y + (this.customSizeHeight || this.height);
```
Thomaash, line 110: Doesn't `??` make more sense here?

lib/network/modules/components/nodes/util/ShapeBase.js
yotamberk: what do you mean?

yotamberk:
As I sent you the question, I realized what I needed to do... Sorry for the fast trigger finger. I've solved the issue

lib/network/modules/components/nodes/shapes/CustomShape.js
```diff
@@ -62,6 +62,11 @@ class CustomShape extends ShapeBase {
       };
     }
 
+    if (drawLater.nodeDimensions) {
+      this.customSizeWidth = drawLater.nodeDimensions.width;
+      this.customSizeHeight = drawLater.nodeDimensions.height;
```
Thomaash, line 67: Isn't `nodeDimenstions` a function?

lib/network/modules/components/nodes/shapes/CustomShape.js
```diff
@@ -1,6 +1,6 @@
-'use strict';
+"use strict";
 
-import ShapeBase from '../util/ShapeBase'
+import ShapeBase from "../util/ShapeBase";
```
Thomaash, line 3: Could you remove the reformatting (it makes git blame much nicer to walk through), please? We can reformat the codebase with Prettier and force consistent formatting, which is probably a very good idea, but this should be done in a separate PR.

lib/network/modules/components/nodes/shapes/CustomShape.js
yotamberk: no, it's actually an object. the docs were wrong. I fixed them

yotamberk:
I think I've solved everything

FlorentMouysset:
Hello everyone,

When I use a custom drawing function, I have the impression that the bounding box is not correct, see the image below.

image


Green is the custom node, red is the associated boundingbox.

Here is a partial code highlighting the problem. I am using version 8.5.6

```typescript
ctxRenderer:
                function ctxRenderer(all: any) {
                    return {
                        drawNode() {
                            all.ctx.fillStyle = 'green'
                            all.ctx.fillRect(all.x, all.y, 100, 100)
                        },
                        nodeDimensions: { width: 100, height: 100 },
                    }
                },

//.....
this.network.on("afterDrawing", (ctx: any) => {
            data.nodes.forEach((node: any) => {
                bb = this.network.getBoundingBox(node.id)
                ctx.fillStyle = 'red'
                ctx.fillRect(bb.left, bb.top, 100, 100)
            })
        })

Thomaash:
Hi @FlorentMouysset,

this is a closed PR, the code has already been merged and therefore no changes can happen here. If you found an issue with version 8.5.6, open a new issue and link this PR, if you think that it's relevant in any way.

Thanks.

PR #1094, opened 2020-10-08
Updating Types from Documentation

### LINKED ISSUE #1093: [Discussion] Types across vis-network and Visjs
Hi,

I'm looking at helping out by adding more types to vis-network (and vis in general) but looking at DefinitelyTyped it's not been updated since May 2019, before vis was deprecated in favour of the community edition vis-network and others. I've also seen the minimal fork for vis-types which looks like is just used to pull any relating commits in from DT.

There's also the ./types in this repo which seems to have newer definitions so my question is where the best place is to submit newer types, (also happy to lend a hand with more migration to TS in general). I don't know if we want the @types/vis to stay at v4.21 as it's deprecated and then a new package (e.g. @types/visjs) for all the new definitions between then and v8.3.2/latest

I had [started a PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/48620) on DT but I've now since closed that for now while I try to gain some guidance.

Let me know your thoughts,

tiberiushunter:
Transferring from PR to DefinitelyTyped
+ Update type definitions for NetworkEvents
+ Updates type definitions for EdgeOptions
+ Refactors Font and ArrowHead typings to reduce repeated types

closes #1093

### DIFF
--- a/types/network/Network.d.ts
+++ b/types/network/Network.d.ts
@@ -895,21 +895,7 @@ export interface NodeOptions {
     y?: boolean,
   };
 
-  font?: string | {
-    color?: string,
-    size?: number, // px
-    face?: string,
-    background?: string,
-    strokeWidth?: number, // px
-    strokeColor?: string,
-    align?: string,
-    vadjust?: number,
-    multi?: boolean | string,
-    bold?: string | FontOptions,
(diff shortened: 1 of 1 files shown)


Thomaash (approved):
LGTM, thanks.

PR #1133, opened 2020-10-25
fix(select): trigger select events for shift+drag selections

### LINKED ISSUE #1128: Selecting with SHIFT key does not trigger select event
Hi!

I've seen that selecting with the shift key (introduced in pr #817 ) is not triggering the `select` event. Is this intended?
Selection event is working fine with control click and long click.

Thanks!

Thomaash:
Selection events were only triggered by clicks but not drags, this PR fixes that.

Some refactoring is also included (selection handling is now encapsulated in a class, some underscore convention fixes and dead code removal…) and E2E test.

Closes #1128.


mojoaxel (approved):
LGTM

yotamberk (approved):

PR #1134, opened 2020-10-25
feat(fit): allow custom zoom level limits

### LINKED ISSUE #1114: Is there no way to give an initial scale value to the Network?
I read the entire documentation and couldn't find a way to properly fit the graph to the container size:

I'd like the network to fit the container. Here it seems too far away and I have to zoom in.

The `fit()` method makes it back to this initial scale.

Is there any way to initially scale up the graph a bit?

Thomaash:
They are optional and default to the same values as before (that is no breaking change), they are limited to values above 0 (zero would lead to infinities, NaNs etc. and I have no idea what negative zoom should mean) and max can't be lower than min. There's visual regression test included.

Closes #1114.


yotamberk (approved):

yotamberk:
does this solve #629 too?

Thomaash:
> does this solve #629 too?

Nope, this applies only to the fit method.

PR #1187, opened 2020-11-18
Optimized isAlphaNumeric

flexo82:
Optimized isAlphaNumeric to use charcodes instead if regex. This gives the function a worst case 60%+ performance on V8 and 40%+ on Spidermonkey


Thomaash (dismissed):
LGTM

Thomaash:
Hi @flexo82,

could you just fix the linting errors? There are `npm run style-fix` and `npm run lint-fix` commands so it should be fairly easy.

Thanks.

Thomaash (approved):
LGTM, thanks.

Issue #535, opened 2020-03-13
Edge labels with border


mladenbrankovic:
Would it be possible to implement the possibility of adding a border around an edge label? Something like the rough sketch I have attached below:

![image](https://user-images.githubusercontent.com/39416660/76622398-b86a5280-6531-11ea-98c0-47d927042fae.png)

I have found [a nearly identical issue](https://github.com/almende/vis/issues/2203) in the old, archived repository, with the only difference being that I'd also need a border radius.

Thomaash:
Hi @mladenbrankovic,

I think this is possible and probably not even particularly difficult to implement. Sadly at the moment there's no maintainer who would have time to implement new features. Would you find the time to work on this? If so that would be great and (not only) I would be very grateful.

mladenbrankovic:
Hi @Thomaash 
I have barely any experience with contributing to open-source projects, but I'll definitely give it a try.

mladenbrankovic:
Hi @Thomaash 

Sorry to bother you again, but I'm having some trouble setting up the package repository locally. When trying to build or test, I get following error:

```
[!] AssertionError: There has to be a single entry file (entry-esnext.{js,ts}) for the ESNext build: expected [] to have a length of 1 but got 0
AssertionError: There has to be a single entry file (entry-esnext.{js,ts}) for the ESNext build: expected [] to have a length of 1 but got 0
    at callback (C:\devsbb\source\open-source\vis-network\node_modules\vis-dev-utils\src\module\generate-rollup-configuration\index.ts:410:18)
    at validate (C:\devsbb\source\open-source\vis-network\node_modules\vis-dev-utils\src\module\generate-rollup-configuration\index.ts:152:12)
    at C:\devsbb\source\open-source\vis-network\node_modules\vis-dev-utils\src\module\generate-rollup-configuration\index.ts:405:5
    at Array.map (<anonymous>)
    at Object.generateRollupConfiguration (C:\devsbb\source\open-source\vis-network\node_modules\vis-dev-utils\src\module\generate-rollup-configuration\index.ts:397:53)
    at Object.<anonymous> (C:\devsbb\source\open-source\vis-network\rollup.build.js:10:32)
    at Module._compile (internal/modules/cjs/loader.js:689:30)
    at Object.require.extensions.(anonymous function) [as .js] (C:\devsbb\source\open-source\vis-network\node_modules\rollup\dist\bin\rollup:1809:24)
[...]

PR #1290, opened 2021-01-18
Edge labels with border and spacing

mtutic:
I found some time to work on #535. With `font.borderColor`, `font.borderWidth` and `font.padding` options, it is possible to add a border and space to the label. Unfortunatly, I didn't take care of the border radius.

I also don't know it this is the right way to do this but I tried it and it works.
Let me know if something is wrong and needs to be corrected!

### DIFF
--- a/lib/network/modules/EdgesHandler.js
+++ b/lib/network/modules/EdgesHandler.js
@@ -56,6 +56,9 @@ class EdgesHandler {
         size: 14, // px
         face: "arial",
         background: "none",
+        borderColor: "none",
+        borderWidth: 0, // px
+        padding: 0, // px
         strokeWidth: 2, // px
         strokeColor: "#ffffff",
         align: "horizontal",
--- a/lib/network/modules/NodesHandler.js
+++ b/lib/network/modules/NodesHandler.js
@@ -59,6 +59,9 @@ class NodesHandler {
         size: 14, // px
         face: "arial",
         background: "none",
+        borderColor: "none",
+        borderWidth: 0, // px
+        padding: 0, // px
         strokeWidth: 0, // px
         strokeColor: "#ffffff",
         align: "center",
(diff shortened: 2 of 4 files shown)


Thomaash (changes requested):
Hi there,

first of all thanks for your work, however there are a few things that need to be done in order to have this merged:

- [ ] There has to be some documentation for this.
- [ ] It should be tested in some way (we have unit, function E2E and visual E2E already set up; pick whichever suits this the most and make sure it works as it should).
- [ ] I think that padding should be a part of size (that's how CSS does it at the very least). The way it is now the edge can be covered by the border or the border can even extend to the other side of the edge. That's most likely not what most users would expect, especially if there was a background to the label as the label would be perceived as a single block and being misplaced (too close to the edge).
- [ ] The border is sometimes rendered in the middle of nowhere and not around the label. I couldn't figure out why but I got into this situation by modifying the basic usage example with the following code:
```typescript
const edges = new vis.DataSet([
  { from: 1, to: 3, label: "EDGE" },
  { from: 1, to: 2, label: "EDGE" },
  { from: 2, to: 4, label: "EDGE" },
  { from: 2, to: 5, label: "EDGE" },
  { from: 3, to: 3, label: "EDGE" },
]);

const options = {
  edges: {
    font: {
      borderColor: "#ACDC00",
      borderWidth: 4,
      padding: 8,
      align: "top",
    },
  },
};
```
- [ ] Another thing missing is validation. If I pass something like:
```typescript
const options = {
  edges: {
    font: {
      borderColor: "Řecko",
      borderWidth: -44,
      padding: Number.NaN,
    },
  },
};
```
I should get a nice and explanatory message of what is wrong, not be left wondering why is nothing happening (in real use case these values could come from botched math expression, user inputs etc.).
- [ ] Lastly, could you just fix the linting errors prior to opening a PR please? It should be as easy as running `npm run style-fix && npm run lint-fix`.

kosirm:
Anything new? Is this going to be added? I need this urgently!

PR #1378, opened 2021-03-06
chore(docs): update HTML in titles references and add an example

Thomaash:
Clarify how to use HTML in titles as some people were confused by the stale docs (from before we switched to innerText due to XSS) and didn't realise it was simple to get the old behavior back.

### DIFF
--- a/docs/network/edges.html
+++ b/docs/network/edges.html
@@ -1516,10 +1516,16 @@ var options: {
         </tr>
         <tr>
           <td>title</td>
-          <td>String</td>
+          <td>String or Element</td>
           <td><code>undefined</code></td>
           <td>
-            The title is shown in a pop-up when the mouse moves over the edge.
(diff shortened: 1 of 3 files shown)


yotamberk (approved):

lamclennan:
I used the following with jQuery to enable HTML in the tool tips from some ajax driven titles. Not sure if there is a more elegant way to do this. The doco didn't strictly help in this case.
```
    //function to display html in tooltips
    var tooltipprevious = null;
    var tooltipflipflop = true;
    $("div").on('DOMSubtreeModified', ".vis-tooltip", function () {
      if ($(this).is(":visible") && tooltipprevious != $(this).html() && tooltipflipflop) {
        var tooltiptext = $(this).text();
        tooltipflipflop = false;
        $(this).html(tooltiptext);
        tooltipprevious = $(this).html();
        tooltipflipflop = true;
      };
    });
```

PR #1530, opened 2021-06-21
Fix custom shape size

cristianoccazinsp:
**Thank you for contributing to vis.js!!**

Please make sure to check the following requirements before creating a pull request:

Simple attempt to fix https://github.com/visjs/vis-network/issues/1494

Additionally, custom shapes can now receive the options object in the renderer to further allow customization and external values.

Lastly, added a new option to allow nodes to be tapped / selected beyond their dimensions.  This is to support being able to select small nodes by adding a small padding to them (defaulting to).  Such a small change that another PR didn't make much sense.

### DIFF
--- a/docs-kr/network/nodes.html
+++ b/docs-kr/network/nodes.html
@@ -279,6 +279,7 @@ var options = {
       coordinateOrigin: 'center'  // only for image and circularImage shapes
     }
     size: 25,
+    touchPadding: 0,
     title: undefined,
     value: undefined,
     widthConstraint: false,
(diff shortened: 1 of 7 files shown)


Thomaash:
Trying to slip a feature in a fix is a really bad practice, atop of that you're exposing internal, undocumented object to the users. If this is something that is really necessary, a proper stable API has to be established and documented first. Anyway, there should be two PRs for this.

cristianoccazinsp:
Sorry I didn’t have time to set up and document a whole new API for the changes, they seemed pretty small. If you think this would still be useful, I can try to put a proper PR for it when I have time, but we needed the change right away for what we are doing.

Weax (commented):
Please, create a single PR for this change..

PR #1533, opened 2021-06-28
Docs: fix typo in examples with inheritColor option for Gephi import

psanan:
This fixes what I think is a typo in the documentation, an extra "s" added to the key `'inheritColor`).

There are a couple of mentions of "inheritColors" in `HISTORY.md` which I left alone, since I don't know if that was a typo at the time, and in any case it's probably not good to rewrite those logs.

Aside: in making this PR I ran across some outdated docs on how to contribute, which could be removed or updated: there are a couple of broken links at the top of `CONTRIBUTING.md`, and the corresponding documents in `misc/` seem out of date (e.g. they mention that PRs should go to a `develop` branch).


Thomaash (approved):
:+1:

Issue #1536, opened 2021-07-01
Network auto focus on hover when bindToWindow is false


msarsha:
When `options.keyboard.bindToWindow = false` and the network element is just outside view port - hovering the element will cause the viewport to scroll to the element (focusing the canvas).

This behaviour is not always required as it can cause "glitches" if a user accidentally hovers on the element.

I created a PR that adds a `skipHighlight` option to allow consumers to opt out from this behaviour.

Example of said behaviour: https://jsbin.com/ginolifusi/edit?html,output

msarsha:
@eilonudi

PR #1537, opened 2021-07-01
feat(interaction-handler): add the ability to skip highlight when bin…

msarsha:
Issue: #1536 

This PR adds the ability to skip auto focus when bindToWindow is false to prevent glitches described on #1536


Thomaash (changes requested):
Could you just rename it to `autoFocus` which would default to `true`? I don't like the idea of disabling auto focus by `skipHighlight` (seems quite cryptic/confusing). Otherwise :+1:.

msarsha (commented):
Done.

Thomaash (approved):

PR #1691, opened 2022-02-01
Prevent reset the node color to the group color from storePositions function

veredwpsemantix:
This pull request fixes the reset of node color when we call to storePositions function.

Example:
before storePositions:
Screenshot1

after storePositions:
Screenshot2

### DIFF
--- a/lib/network/modules/components/Node.js
+++ b/lib/network/modules/components/Node.js
@@ -323,7 +323,7 @@ class Node {
       (p) => newOptions[p] != null
     );
     // Always skip merging group font options into parent; these are required to be distinct for labels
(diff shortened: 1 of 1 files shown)


Thomaash:
Hi @veredwpsemantix,

could you provide an MWE of the issue please? From what you posted this seems like a pretty serious issue in vis-data and should therefore be fixed there. What you provided is more of a hack around it (assuming I understand the problem correctly, which is difficult to gauge without an MWE).

Thanks and sorry for the atrocious response time

PS: There are JSFiddle and CodePen buttons on https://visjs.github.io/vis-network/examples/ to make MWE creation easier.

PR #1972, opened 2023-07-13
Handle both mouse wheel and trackpad

### LINKED ISSUE #1882: Allow separate zoom speeds for wheel and trackpad
Prior to [this commit](https://github.com/visjs/vis-network/commit/aa7e6646acbb0ea52bf5a7041fec8e103f8dfab5), vis-network worked well for our trackpad users and mousewheel users - zoom speed felt appropriate for both. Now, however, we're getting feedback that the speed is either too fast (from trackpad users) or too slow (from mousewheel users).

We could add a setting to our app for zoom speed, but we'd rather have the relationship between wheel speed and trackpad speed be fixed upstream. Two options appear to be:

1. Detect which one is being used for scrolling and then adjust the zoom by some fixed coefficient
2. Allow separate `zoomSpeed` settings for the two different scroll types.

TedDriggs:
The current code ignores magnitude differences in deltaY, resulting in an experience that is either too fast on trackpad or too slow with a mousewheel.

This change allows those to differ within a clamped range, so that one zoomSpeed value works well for both input devices.

Fixes #1882


jeffrey-e:
Looking forward to this improvement!

TedDriggs:
@Thomaash could you take a look at this PR? Having a fix for this would let us remove a very clunky workaround in our app where we have to give the user a setting to tell us which input device they're using.

PR #2006, opened 2023-10-04
Fix typo in layout.html documentation

harryrichman:
The documentation seemed to have a simple mistake, where `leaves` is the default parameter rather than `roots`.


Thomaash (approved):
thanks

PR #2024, opened 2023-10-13
fix(horizontal-strategy): add missing sort method

schelmo:
`.sort()` was missing in HorizontalStrategy in #1946


Thomaash (approved):
thanks

PR #2219, opened 2025-05-13
fix(hierarchical): compress level values to eliminate gaps in cyclic graphs

beatriceev:
### Proposal: Compact level values to improve layout consistency in cyclic graphs

When working with graphs that include cycles, the fallback layout (`fillLevelsByDirectionCyclic`) can sometimes assign sparse or widely spaced level values (e.g., levels like 0, 1, 50). This causes nodes to be unevenly spaced in hierarchical mode, especially in partially cyclic structures.

This PR introduces a small enhancement to `setMinLevelToZero()`:

- After the usual min-level normalization, it remaps the level values to a contiguous sequence (e.g., 0, 1, 50 → 0, 1, 2).

- This helps produce a more balanced layout without affecting node relationships or acyclic graphs.

The change is minimal, and improves layout consistency in edge cases. Happy to discuss if there’s a better place for this logic.

**Demo**: https://jsfiddle.net/uvLcg98k/21/
> Comment out line 84: `network.on('afterDrawing', () => patchLayoutEngine());` to compare with the original behavior.


beatriceev:
Update: Fixed an edge case where `fillLevelsByDirection` would assign all nodes to level 0 if no entry nodes (i.e., roots or leaves, depending on the `shakeTowards` configuration) were detected — such as in a cyclic graph. It now correctly falls back to `fillLevelsByDirectionCyclic` in that scenario. 

Without this fallback, all nodes would be placed on level 0, which can cause the graph to render in an unintended direction. For example, if the user specifies direction: "LR", the result may visually resemble a "UD" layout due to all nodes being on the same level.

lib/network/modules/layout-engine/index.ts
```diff
@@ -50,19 +49,31 @@ function fillLevelsByDirectionCyclic(
...
+
   return fillLevelsByDirection(
     // Pick only leaves (nodes without children).
-    (node): boolean =>
-      node.edges
-        // Take only visible nodes into account.
-        .filter((edge): boolean => nodes.has(edge.toId))
-        // Check that all edges lead to this node (leaf).
-        .every((edge): boolean => edge.to === node),
+    (node): boolean => entryNodes.includes(node),
```
Thomaash, line 76: What is the purpose of this? Won't the condition at the end catch cyclic graphs every time this would? It seems to me like just added overhead for people who are not abusing this layout for cyclic graphs. Off course I may be missing something.

lib/network/modules/layout-engine/index.ts
```diff
@@ -73,19 +84,31 @@ export function fillLevelsByDirectionLeaves(nodes: Map<Id, Node>): Levels {
...
+
   return fillLevelsByDirection(
     // Pick only roots (nodes without parents).
-    (node): boolean =>
-      node.edges
-        // Take only visible nodes into account.
-        .filter((edge): boolean => nodes.has(edge.toId))
-        // Check that all edges lead from this node (root).
-        .every((edge): boolean => edge.from === node),
+    (node): boolean => entryNodes.includes(node),
```
Thomaash, line 111: What is the purpose of this? Won't the condition at the end catch cyclic graphs every time this would? It seems to me like just added overhead for people who are not abusing this layout for cyclic graphs. Off course I may be missing something.

lib/network/modules/LayoutEngine.js
```diff
@@ -213,11 +206,27 @@ class HierarchicalStatus {
         }
       }
     }
+
+    // Added: Compress level values to eliminate gaps (for cyclic)
+    const remap = new Map();
+    let newLevel = 0;
+    const uniqueSortedLevels = [
+      ...new Set(Object.values(this.levels).sort((a, b) => a - b)),
```
Thomaash, line 214: Just a performance tweak: Sort after you run it through Set. Set doesn't care about order and there will be fewer elements to sort once only unique are left.

lib/network/modules/LayoutEngine.js
```diff
@@ -213,11 +206,27 @@ class HierarchicalStatus {
         }
       }
     }
+
+    // Added: Compress level values to eliminate gaps (for cyclic)
```
Thomaash, line 210: Don't put added into comments. I can see what is added in the diff and once merged, it will stay in the code forever.

lib/network/modules/LayoutEngine.js
```diff
@@ -213,11 +206,27 @@ class HierarchicalStatus {
         }
       }
     }
+
```
Thomaash, line 209: Isn't the part above now redundant?

Thomaash (changes requested):
Was your intent to only submit the `// Added: Compress level values to eliminate gaps (for cyclic)` part? If so, then just delete all the other changes around, address the 3 comments I have there (should be easy) and it can be merged.

The rest doesn't really make sense to me. If you wan't to keep that in the PR, I'll need a detailed explanation as to why it's useful.

Either way, thanks for your contribution, it's always great to see people trying to fix the issues they encounter.

lib/network/modules/LayoutEngine.js
```diff
@@ -49,7 +49,6 @@ import {
 
 /**
  * Container for derived data on current network, relating to hierarchy.
- *
```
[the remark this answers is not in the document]
Thomaash: This happened because of JSDoc ESLint plugin update. Rebase and don't worry about it.

lib/network/modules/LayoutEngine.js
```diff
@@ -1614,8 +1601,7 @@ class LayoutEngine {
 
   /**
    * Crawl over the entire network and use a callback on each node couple that is connected to each other.
-   *
-   * @param {Function} [callback=function(){}]          | will receive nodeA, nodeB and the connecting edge. A and B are distinct.
+   * @param {Function} [callback]          | will receive nodeA, nodeB and the connecting edge. A and B are distinct.
```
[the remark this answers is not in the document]
Thomaash: This happened because of JSDoc ESLint plugin update. Rebase and don't worry about it.

It's actually superfluous and error prone since it it duplicated using the ES6 default value syntax a few lines below.

lib/network/modules/layout-engine/index.ts
beatriceev: Hi Tomáš,

I understand your point; there are definitely many cases where the graph has no leaves but is also **not strictly cyclic** (e.g., mutual connections, self-loops, or dense interlinked clusters). So yes, the current fallback could be unnecessary overhead in those cases.

Given that, do you think it would make sense to explicitly check whether the graph is fully cyclic before falling back? Or is it acceptable that cyclic graphs don't follow the `shakeTowards` configuration?

The reason I ask is that without this fallback, a fully cyclic graph (i.e., no leaves at all) would result in all nodes being placed on level 0. This can cause the layout to render in an unintended direction. For example, if the user specifies a horizontal layout, the result might visually resemble a vertical one due to all nodes being aligned at the same level.

lib/network/modules/LayoutEngine.js
beatriceev: Resolved!

lib/network/modules/LayoutEngine.js
beatriceev: Applied the suggested change: deduplication now happens before sorting to improve performance. Thank you!

Thomaash (approved):
Thanks for you contribution.

PS: I made just some small changes in order not to rename the project and removed the now superfluous code.

PR #2430, opened 2026-05-14
Allow wind to be a function of nodeId

Zeophlite:
Allows for the physics simulation to give per-node wind, rather than it being fixed for every node

Screenshot 2026-05-14 at 7 51 52 pm


Zeophlite:
Closes https://github.com/visjs/vis-network/issues/2085

Thomaash (approved):
Looks good to me, only one thing for next time, please don't remove examples of old behavior that is still relevant, thanks

PR #2444, opened 2026-06-01
fix(node.icon): Allow Node.icon to inherit Node.size

### LINKED ISSUE #385: Change selected node styles
I have groups of nodes, and i am using `shape: icon`
i was trying to update the selected node, size, color, shadow and few more things. 

i found this example which is really helpful 
https://github.com/almende/vis/blob/master/examples/network/other/chosen.html

but in the example, when i am trying to change the node size it doesn't work as long as i am using the shape: icon

```
color: "#97C2FC"
borderWidth: 20
borderColor: "#2B7CE9"
size: 100
borderDashes: false
borderRadius: 6
shadow: true
shadowColor: "rgba(0,0,0,0.5)"
shadowSize: 50
shadowX: 0
shadowY: 0
```

the values.size is updated but it doesn't really reflect on the network itself. 

I wonder if there is any other way to be able to update a node. 
I tried ` nodes.updateOnly(id, props) `
it works well, but this cause the network to be updated whenever i select a node. 
and this makes my network to be destroyed and recreated and i don't expect this behaviour

hwelch-fle:
*No AI used at any stage of this PR*

Change option resolution order for Node to allow node icon size to inherit from node size

I tried to keep the resolution simple. Since `icon.size` is the only shape that uses a seperate `size` property, I had to unset the defaults in `NodesHandler` for both `size` and `icon.size`.

These options are resolved in `NodeBase.setOptions` now.

Old Icon Size Resolution:

`<nodeobj>.icon.size -> <groupobj>.icon.size -> <network.nodes>.icon.size`

New Icon Size Resolution:

```
(node->group->network.nodes).icon.size
--then--
(node->group->network.nodes).size
--then--
50
```

I'm not sure if it's better to have the size be inherited this way though:
```
node.icon.size -> node.size
--then--
group.icon.size -> group.size
--then--
network.icon.size -> network.size
--then--
50 
```

If the second option is more logical I will gladly ammend this PR to do it that way. I just figured that `icon.size` being directly set at any point should probably supercede the `size` fallbacks to prevent issues with existing setups.

This might also cause breaking changes if someone is setting `size` and leaving it set on an icon node because they saw no change. The fix would be to either remove the `size` key from the node or to set `icon.size` at any level of the config.

resolves #385

### DIFF
--- a/docs/network/nodes.html
+++ b/docs/network/nodes.html
@@ -1111,7 +1111,10 @@ function({ ctx, x, y, state: { selected, hover }, style }) {
           <td class="indent">icon.size</td>
           <td>Number</td>
           <td><code>50</code></td>
-          <td>The size of the icon.</td>
+          <td>
+            The size of the icon. (will inherit <code>size</code> if
+            <code>undefined</code>)
+          </td>
         </tr>
         <tr parent="icon" class="hidden">
           <td class="indent">icon.color</td>
--- a/docs/network/nodes.html
+++ b/docs/network/nodes.html
@@ -1111,9 +1111,22 @@ function({ ctx, x, y, state: { selected, hover }, style }) {
           <td class="indent">icon.size</td>
           <td>Number</td>
           <td><code>50</code></td>
+          <td>The size of the icon.</td>
+        </tr>
+        <tr parent="icon" class="hidden">
+          <td class="indent">icon.inheritSize</td>
+          <td>Boolean</td>
+          <td><code>false</code></td>
           <td>
-            The size of the icon. (will inherit <code>size</code> if
-            <code>undefined</code>)
+            When <code>true</code>, inherit <code>size</code> from
+            <code>node.size</code>. Takes precedence over
+            <code>icon.size</code>. <br /><i
+              >Note: This flag exists to preserve the original behavior of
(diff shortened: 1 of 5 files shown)


hwelch-fle:
It does seem that two of the visual regression checks are failing, but that's expected since those have a `size` set on the node level with `icon.size` left undefined.

If we want to add this, but prevent regressions in networks created with the size unaware icons, we could maybe add a new `inherit_size` option to the ~~network~~ node icon options?

hwelch-fle:
Additional commits should preserve original behavior and hide size inheriting behind a new `inheritSize` property.

Value add here is that currently, icons that are set in groups can't be dynamically sized without directly copying the group icon object to the node and updating the size.

This will allow you to use icons and drive the size the same way you drive the size of every other shape type.

[Here's a sample graph that this would benefit](https://hwelch-fle.github.io/pygraph/)

I wanted to set the icon groups based on file extension, then size the nodes based on line count, but the group icon size is set to 50 and overwrites the node size currently.

hwelch-fle:
If this is something that will be merged, we might want to do the same thing with `icon.color` too? I will also make those changes if you want.

PR #2450, opened 2026-06-19
fix(physics): stop forceAtlas2Based layout from rotating forever (#2193)

### LINKED ISSUE #2193: Graph layout converges but graph keeps rotating around center of mass
I have a graph which looks like this:

image

And while the general layout converges quickly, the whole "wheel" keeps rotating forever. This is reproducable, independent of the randomized initial layout. However it only happens with the atlas2 based layout (which seems to give the nicest results in principle ...)

It's using the following parameters:

```
        'physics': {
            "forceAtlas2Based": { "springLength": 100 },
            "minVelocity": 0.75,
            "solver": "forceAtlas2Based",
            'stabilization': { 'iterations': 20 }
        }
```

This is with visdcc 0.0.50 (not clear to me which visjs version that is).

Is this known behavior? Is there a workaround? (except avoiding atlas2 ...)

bojanstef:
## Summary

Fixes #2193. With `forceAtlas2Based`, an asymmetric graph's layout converges in shape but then keeps **rotating about its center of mass forever**. No other solver does this.

## Root cause

`FA2BasedRepulsionSolver` scaled the pairwise repulsion by the **receiving** node's degree only (`node.edges.length + 1`). So the force on *i* from *j* is ∝ `deg_i` while the force on *j* from *i* is ∝ `deg_j` — non-reciprocal whenever the degrees differ, violating Newton's third law. That injects a small net torque every tick; velocity damping caps the resulting spin but never removes it, so the converged layout rotates at a constant terminal rate. (On a symmetric graph the per-pair torques cancel, which is why it went unnoticed.)

## Fix

Keep ForceAtlas2's degree-weighted repulsion — it's what pushes high-degree hubs to the periphery — but apply it **symmetrically**, as in the original FA2 formulation: repulsion ∝ `(deg_i + 1)(deg_j + 1)`. This is done by folding `(degree + 1)` into the Barnes–Hut **effective mass** used both to build the tree and to evaluate the force, so the pairwise force is reciprocal again and torque-free. Hubs are still pushed outward; the rotation is gone.

## Tests

Rewrote `test/physics/forceatlas2-rotation.test.ts` to measure the **actual rotation of node positions** (best-fit/Procrustes angle over a long, never-stabilizing run) rather than a velocity proxy. Deterministic (seeded positions, fixed timestep). Asserts net rotation < 5° (bug ≈ 122°, fix ≈ 0.4°); red on `master`, green with the fix.

Validated across hub, barbell, scale-free and random graphs: net rotation drops **7×–5800×** and the layout settles in every case, while degree-based hub spreading is preserved.

## Snapshots

The fix changes converged FA2 layouts, so the 6 `forceAtlas2Based` snapshots whose graphs have edges are regenerated. The 2-disconnected FA2 snapshot is unchanged (degree-1 nodes), and no other solver's snapshots change.


bojanstef:
@Zeophlite @Thomaash let me know what you think

Thomaash:
Hi @bojanstef, first of all thanks for contributing. However, AFAIK degree based repulsion is the selling point of ForceAtlas2 (it pushes highly interconnected hubs to the periphery, making them easier to discern), removing it seems as a bad idea to me.

bojanstef:
@Thomaash ah, understood. I was resolving a symptom and not the cause. Let me take another stab at it...

